PR #11, opened 2011-10-08
make sure we build a 64 bits enabled platform on Darwin

benoitc:
This patch detect if 64 bits is supported and build erlang accordingly.


mrallen1:
This PR has been open for >4 years. I'm going to close it - if it's still a problem, please open a new issue.

PR #18, opened 2011-12-31
Make kerl work with OTP < R15B on newer Solarii

timclassic:
Tested on OpenIndiana oi_151a.  The embedded patch to inet_drv.c has been tested on Solaris 10 and newer, though GNU sed is necessary for kerl to function overall.

### DIFF
--- a/kerl
+++ b/kerl
@@ -288,6 +288,20 @@ do_git_build()
 
 do_build()
 {
+    assert_valid_release $1
+    assert_build_name_unused $2
+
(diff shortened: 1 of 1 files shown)


mrallen1:
Closing this because it's been incorporated into #138.

mrallen1:
@timclassic I know it's been forever and a day, but your PR was merged in #138. If you're still using Solaris and Erlang, it'd be awesome if you could test it. Thank you.

PR #43, opened 2013-03-20
Do not clear ENV config vars

yfyf:
According to the docs it is possible to configure kerl both by using `.kerlrc`
and passing ENV variables from the command-line, i.e.

    KERL_INSTALL_MANPAGES=yes ./kerl install <build> <path>

However the ENV vars were being cleared/overriden, so you could only use `.kerlrc`.

This fixes it.

### DIFF
--- a/kerl
+++ b/kerl
@@ -23,11 +23,13 @@
 
 ERLANG_DOWNLOAD_URL=http://www.erlang.org/download
 
-KERL_BASE_DIR="$HOME/.kerl"
-KERL_CONFIG="$HOME/.kerlrc"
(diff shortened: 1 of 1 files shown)


mrallen1:
Let's get this included after we figure out what the merge conflict is.

sanmiguel:
I have now rebased this branch, and fixed a related bug (2 more env vars were missed when backing up after reading `$KERL_CONFIG`) on top.

I shall close this PR and submit a new one from that rebased branch.

@yfyf thanks for the contribution!

PR #46, opened 2013-06-16
make _get_comp_words_by_ref more strong

mryufeng:


mrallen1:
Seems like this patch ensures that the `_get_comp_words_by_ref` function exists instead of blindly assuming it exists.

PR #77, opened 2014-06-11
Add submodule recursive init/update after branch checkout in git build

sanmiguel:
SD-erlang OTP fork (https://github.com/release-project/otp) uses a submodule to bring in `percept2` application. This causes the build to fail.

This PR adds a (possibly naive) `git submodule update --init --recursive` **after** checking out the required branch

### DIFF
--- a/kerl
+++ b/kerl
@@ -283,6 +283,12 @@ do_git_build()
         rm -Rf "$KERL_BUILD_DIR/$3"
         exit 1
     fi
+    git submodule update --init --recursive > /dev/null 2>&1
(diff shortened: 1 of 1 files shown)


mrallen1:
Ohai! What should we do with this PR?

sanmiguel:
Good question: I haven't actually touched SD-erlang since submitting this PR and it doesn't look like it's been touched since Dec '14.

I think we should probably just shelve this for now.

mrallen1:
I'm going to close it then.

PR #78, opened 2014-06-12
move sed to a var and use gsed on OpenBSD

qbit:
OpenBSD's sed doesn't have -i. This diff splits it out into a var so it can be changed as needed.


qbit:
OpenBSD has a -i now, I am going to close this :D

PR #85, opened 2014-10-23
support "export KERL_BASE_DIR=/path/to/KERL_BASE_DIR # default $HOME/.kerl"

cola1129:


mrallen1:
Thanks for the patch. I don't understand the intent of this PR - can you help me understand the rationale behind the change?

sanmiguel:
I think this might have been addressed by #116 - @cola1129 can you confirm that what you wanted was to be able to do something like:

```
$ export KERL_BASE_DIR="/some/path/to/somewhere"
$ kerl build ...
```
and have `kerl` use `/some/path/to/somewhere` as it's base directory?

If so, that should definitely now be possible after #116

mrallen1:
I'm going to close this since we haven't heard anything further. Please open a new ticket if needed.

PR #92, opened 2015-01-07
Fix Darwin LLVM detection

orien:
LLVM is now specified with CAPS. We should search in a non-case sensitive way (`grep -i`).

Also prevent printing to STDERR when checking the version.


yrashk:
Thanks!

PR #97, opened 2015-06-24
Escape angle brackets

legoscia:
This way, the placeholders show up in the HTML output.


mrallen1:
We just merged #105 which addresses this issue too. Thanks.

PR #101, opened 2015-07-03
Build a Dialyzer PLT when installing a build

erszcz:
It's convenient to have a global PLT with all Erlang/OTP apps to reuse in multiple projects and cut down on time needed to build a per-project PLT.
This PR makes it possible to define a `KERL_BUILD_PLT` (on the command line or in `.kerlrc`) variable, which if set will make `kerl` build a PLT of all apps going into an installation. `kerl plt` will print availability of a PLT for the active installation. Such a modification was also made to `kerl status`.
Tested with 17.1 and 18.0 - no problems encountered.

This is based on top of #100 issued separately.


tomekowal:
:+1:

mrallen1:
Really like this idea. :+1: What do you think @sanmiguel?

sanmiguel:
Agreed @mrallen1 - this seems like something that would be very useful. :+1:

mrallen1:
I think we want to refactor how documentation get built even further, but this is a nice 1st step.

PR #110, opened 2016-02-03
Auto add OpenSSL path on El Capitan

mrallen1:
Apple officially removed OpenSSL libraries from El Capitan, so Erlang builds that rely on Xcode (which used to work) will now be built without functional crypto applications.  This patch auto-adds the header/library location so that Erlang releases will build with functional crypto on El Capitan.

There's probably a better way to do this but this seems to work ok.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@ kerl
 
 Easy building and installing of [Erlang/OTP](http://www.erlang.org) instances
 
-Kerl aims to be shell agnostic and its only dependencies, excluding what's required to actually build Erlang/OTP, are curl and git.
+Kerl aims to be shell agnostic and its only dependencies, excluding what's required to actually build Erlang/OTP, are `curl` and `git`.
 
 All is done so that, once a specific release has been built, creating a new installation is as fast as possible.
 
@@ -64,7 +64,7 @@ Note that named builds allow you to have different builds for the same Erlang/OT
     Building Erlang/OTP R14B02 (r14b02_hipe), please wait...
     Erlang/OTP R14B02 (r14b02_hipe) has been successfully built
 
-(Note that kerl uses the otp_build script internally, and './otp_build configure' disables HiPE on linux)
+(Note that kerl uses the otp_build script internally, and `./otp_build configure` disables HiPE on linux)
 
 You can verify your build has been registered:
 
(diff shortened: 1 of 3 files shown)


tburghart:
Suggestions:

Confirm that there are no conflicting SSL options in `KERL_CONFIGURE_OPTIONS` before adding this.

Confirm that `/usr/include/openssl` and /usr/local/include/openssl` are not already present - many developers amy have already added them.

Starting with OTP-17, the `--with-ssl-incl=` option is preferable, as it doesn't affect the library search path.

mrallen1:
What do you think about this revision @tburghart?

tburghart:
More comments (aren't you glad you asked?):

The current PR only applies to non-Git builds, but it really should run in Git and non-Git builds equally.

I don't *think* the compiler will look in `/usr/local/include/openssl` by default on OS X, so the user would have had to have set `--with-ssl=/usr/local` to look there anyway ... though I could be wrong.

Also, OTP-17 added the `--with-ssl-incl` switch to look in the specified path *only* for the headers and assume the libraries themselves were still on the system path, which is what you want given the choice. I don't see an obvious source of the target release (the `$otp_release` variable in my code below) at the point when the configure options are currently being calculated, but the presence of the file `$ERL_TOP/OTP_VERSION` coincides with OTP-17 as a cheap-n-dirty version check if the calculation is moved to after the source tree is downloaded.

With that, I believe the logic should be:

```bash
if echo -n "$KERL_CONFIGURE_OPTIONS" | egrep -qvw 'ssl' \
&& [ "$KERL_SYSTEM" == 'Darwin' ] \
&& [ `uname -r | cut -d. -f1` -gt 14 ] \
&& [ ! -d '/usr/include/openssl' ]
then
    xc_ssl='/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/swift-migrator/sdk/MacOSX.sdk/usr'
    if [ -d "$xc_ssl/include/openssl" ]
    then
        if [ $otp_release -ge 17 ]
        then
            KERL_CONFIGURE_OPTIONS="$KERL_CONFIGURE_OPTIONS --with-ssl --with-ssl-incl=$xc_ssl"
        else
            KERL_CONFIGURE_OPTIONS="$KERL_CONFIGURE_OPTIONS --with-ssl=$xc_ssl"
        fi
    else
        echo 'WARNING: No OpenSSL library was found in the usual places. Your Erlang will be built without crypto support!'
    fi
    unset xc_ssl
fi
```

vat-sqor:
is anyone working on getting this, or something like it merged? 
i spent like an hour today trying to build erlang w/ kerl before finding this PR :(

mrallen1:
I'm definitely working on it, yep. :) I just haven't incorporated @tburghart's latest suggestion into this patch. 

As far as merging it, unless @yrashk approves or hands out commit bits not sure what else can be done. It's definitely a pretty irritating situation though.

mrallen1:
OK, rebased and ready for review. I added a shell function to pull out the OTP release from the build request. This only works for tarball builds right now because git tags can be anything.

mrallen1:
Ugh. This isn't what I wanted at all. Going to close this and start a new one.

PR #115, opened 2016-04-11
Use dot instead of source

ten0s:
In Ubuntu /bin/sh is /bin/dash, which doesn't support the source command.


mrallen1:
This got changed when we merged #106 - thanks.

PR #116, opened 2016-04-13
Allow env vars to override default values

sanmiguel:
TPRFKA #43 

Thanks to @yfyf for the contribution.

I'd like to come back to this at a later date - it feels like there ought to be a cleaner way to manage these variables coming from 3 different places (current env, `$KERL_CONFIG` and baked-in defaults).

In the meantime, I think this covers us well enough.


mrallen1:
:+1:

PR #117, opened 2016-04-13
If using OS X El Capitan, try to add OpenSSL

mrallen1:
Added a shell function to return the OTP release from tarball builds.

Added logic to try and find OpenSSL libraries in the usual places. If we can't find any emit a warning.


mrallen1:
Supecedes #110

sanmiguel:
:+1:

Issue #120, opened 2016-04-19
Kerl fails with "No newest valid release"


ooshlablu:
Installing any (Basho at least) erlang fails when using the kerl version from HEAD with the error message: "No newest valid release". The bizarre thing is that erlang gets built and installed just fine; kerl just exits with that error no matter what happens. Kerl 0.8 works as expected.

I have reproduced with the following erlang installs from git (https://github.com/basho/otp.git):
* R16B02-basho10 (git tag: OTP_R16B02_basho10)
* R16B02-basho8 (git tag: OTP_R16B02_basho8)
* R15B01 (git tag: basho_OTP_R15B01p)

I have tested on the following (64 bit) OS's:
* Ubuntu Precise
* Ubuntu Trusty
* Debian Jessie
* Debian Wheezy
* CentOS 6
* CentOS 7

I will be working with @mrallen1 to get him set up with the environment that I used to reproduce the issue.

sanmiguel:
FWIW I can replicate this with kerl 0.9.1 (via homebrew) on OSX 10.10.5, too. 
![screenshot 2016-04-20 01 01 02](https://cloud.githubusercontent.com/assets/2122822/14658369/88e2444c-0693-11e6-831a-f2404063e059.png)

Pretty sure this error message is related to installing docs, as none have shown up in this installation:

```bash
$ erl -man gen_fsm
No manual entry for gen_fsm
```

lucafavatella:
Change triggering this shall be somewhere in [get_newest_valid_release](https://github.com/kerl/kerl/compare/0.8...0.9.1#diff-75743bad0452412b658dabebd6b439f4R219).

lucafavatella:
... or for some reason [this](https://github.com/kerl/kerl/compare/0.8...0.9.1#diff-75743bad0452412b658dabebd6b439f4R144) regular expression does not support the current version of the OTP download [page](http://www.erlang.org/download).

(I am trying to install 18.3.1.)

lucafavatella:
I get this issue running `kerl install OTP-18.3.1 MYPATH` where 18.3.1 built using `kerl build git ...` hence I focussed on get_newest_valid_release code path.

The root cause seems to be that refactoring in 0.9.1 compared to 0.8 broke the semantics of the script. I extracted two minimum code snippets for showing what looks like issue. My otp_releases file is a local one that I had but it is not relevant I believe.

## Kerl 0.8

```
$ cat kerl-0.8-get_newest_valid_release
[...]

Issue #114, opened 2016-04-06
Fix manpage installation


sanmiguel:
We currently grab the docs as a .tar.gz based on the release we're building - fine. But if we're building via `kerl build git ...` we pull docs for *whatever the latest release is*.

Let's use the way `erlbrew` does it, and call `make docs` in the build dir

sanmiguel:
Closed by #122

PR #121, opened 2016-04-20
Address install issues

mrallen1:
Addresses #120 

These changes include the following:
* rewrite `get_newest_valid_release()` to be simpler - the old way set up a while loop but it only ever has a single value.  So let's just put it in a variable and test it.
* Got a build error when I set the `./configure` script to use `--with-ssl` with no path and `--with-ssl-incl`. Just go back to using `--with-ssl=PATH` as before. (Linked to #110)
* If a build fails, tail the last bit of the log file automatically.

I have successfully built and installed 18.3 from a tarball and the basho10 tag from a git build.

### DIFF
--- a/kerl
+++ b/kerl
@@ -422,7 +422,7 @@ do_build()
     fi
     if [ $? -ne 0 ]; then
         echo "Build failed:"
-        tail $LOGFILE
+        tail "$LOGFILE"
         echo
         echo "Please see $LOGFILE for full details."
         list_remove builds "$1 $2"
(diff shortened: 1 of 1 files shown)


sanmiguel:
👍

This works for me on OSX 10.10.5 and Ubuntu 14.04.3.

It's noteworthy that I still haven't done #114 so while the installation from a git repository no longer throws an error when fetching manpages/docs, the actual copy you get is from the **newest** erlang (i.e. 18.3) regardless of which erlang you installed.

Issue #111, opened 2016-02-13
Build OTP from git (was: Missing releases)


tuncer:
Due to looking just in http://erlang.org/download/ for releases, kerl misses many official releases that weren't published as full-fledged ones (with a Windows package, etc.). One way to solve this is download tarballs for OTP* tags like, say, https://github.com/erlang/otp/archive/OTP-18.2.3.tar.gz.

This is also the reason why Travis-CI doesn't provide all official releases, because the prebuilt OTP versions are made with kerl.

The fact that doc archives are meant to be downloaded as well, and how to combine that with downloading git tag archives by default, while there are no matching doc archives, raises more questions.

mrallen1:
Are you proposing that kerl grab point releases from github tag information?

fenollp:
Then how about

```shell
git ls-remote --tags https://github.com/erlang/otp \
    | grep -Eo 'OTP[_-][^^{}]+' \
    | sort -u
```

which gives back 

```
OTP-17.0
OTP-17.0.1
OTP-17.0.2
OTP-17.1
OTP-17.1.1
OTP-17.1.2
OTP-17.2
OTP-17.2.1
OTP-17.2.2
OTP-17.3
OTP-17.3.1
OTP-17.3.2
OTP-17.3.3
OTP-17.3.4
OTP-17.4
OTP-17.4.1
OTP-17.5
OTP-17.5.1
OTP-17.5.2
OTP-17.5.3
OTP-17.5.4
OTP-17.5.5
OTP-17.5.6
OTP-17.5.6.1
OTP-17.5.6.2
OTP-17.5.6.3
OTP-17.5.6.4
OTP-17.5.6.5
OTP-17.5.6.6
OTP-17.5.6.7
OTP-17.5.6.8
OTP-18.0
OTP-18.0-rc1
OTP-18.0-rc2
OTP-18.0.1
OTP-18.0.2
OTP-18.0.3
OTP-18.1
OTP-18.1.1
OTP-18.1.2
OTP-18.1.3
OTP-18.1.4
OTP-18.1.5
OTP-18.2
OTP-18.2.1
OTP-18.2.2
OTP-18.2.3
OTP-18.2.4
OTP-18.2.4.1
OTP-18.3
OTP-18.3.1
OTP_17.0-rc1
OTP_17.0-rc2
OTP_R13B03
OTP_R13B04
OTP_R14A
OTP_R14B
OTP_R14B01
OTP_R14B02
OTP_R14B03
[...]

Issue #68, opened 2014-04-11
Compiling with --enable-native-libs


essen:
Hello,

If you do the following with current kerl it fails to compile everything natively:

```
KERL_CONFIGURE_OPTIONS=--enable-native-libs kerl build 17.0 17.0_native
kerl install 17.0_native ~/erlang/17.0_native
. /home/essen/erlang/17.0_native/activate
% erl                                          
Erlang/OTP 17 [erts-6.0] [source] [64-bit] [smp:4:4] [async-threads:10] [hipe] [kernel-poll:false]

Eshell V6.0  (abort with ^G)
1> code:is_module_native(lists).
false
2> 
```

The reason for this is best explained in this email: http://erlang.org/pipermail/erlang-questions/2014-April/078460.html

Modifying kerl this way made it work as expected:

```
    if [ -n "$KERL_USE_AUTOCONF" ]; then
        ./otp_build autoconf $KERL_CONFIGURE_OPTIONS > "$LOGFILE" 2>&1 && \
            ./otp_build configure $KERL_CONFIGURE_OPTIONS > "$LOGFILE" 2>&1
        make clean
        ./otp_build configure $KERL_CONFIGURE_OPTIONS > "$LOGFILE" 2>&1
    else
        ./otp_build configure $KERL_CONFIGURE_OPTIONS > "$LOGFILE" 2>&1
        make clean
        ./otp_build configure $KERL_CONFIGURE_OPTIONS > "$LOGFILE" 2>&1
    fi
```

However this may or may not be something you guys want. AFAIK it's only needed for native libs so perhaps this can only be done then?

mrallen1:
Good suggestion. Thanks.

mrallen1:
I like the idea of removing the .beam files that come prepackaged with the tarball releases but I am wondering if it would be easier and simpler to have something like `find . -name *.beam -exec rm {}\;` to clean out the pre-existing beam files in the case when `--enable-native-libs` is turned on.

essen:
Not sure, it's been a while. I've since switched to fetching directly from git tags, the problem might also exist there (due to https://github.com/erlang/otp/tree/maint/erts/preloaded/ebin) though it's also been a while and I don't have those builds anymore.

mrallen1:
Should be fully addressed by building OTP from github tarballs, as merged in #122.

essen:
[...]

PR #122, opened 2016-04-25
Support release and tarball retrieval from github

mrallen1:
This PR adds the ability to download releases and tar balls from github instead of erlang.org, as suggested in #111 This has three advantages:

1. Significantly faster download speeds. Not only is github using a CDN but the tarball sizes are roughly half the size of the official Ericsson releases.
2. All of the (reasonably recent) OTP releases can be built (even point releases like 17.5.3.1)
3. Compiling to native libs should succeed because there are no "precompiled" non-native beam files included in the github tar balls. (See #68 for details)

There are also the following disadvantages:

1. Must use autoconf for builds - so there is a new hard dependency on autoconf
2. Must use git (There is a suggestion in #111 to use a HTTP API call but it requires JSON parsing - we already call out a dependency on git in the current README, so it doesn't seem too horrible to assume people will have it for builds.)
3. Really old OTPs are not tagged in the github repo.

This PR retains the ability to download and install "the traditional way" too - but you must set a shell environment variable `$KERL_BUILD_BACKEND=tarball` to get it.  It would probably be smarter to test if `git --version` returns anything and use git, tarball otherwise but that's not how it works right now.

### DIFF
--- a/kerl
+++ b/kerl
@@ -174,13 +174,6 @@ get_tarball_releases()
         sed -e "s/^R\(.*\)/\1:R\1/" | sed -e "s/^\([^\:]*\)$/\1-z:\1/" | sort | cut -d':' -f2
 }
 
-update_git_remote()
-{
-    cd "$KERL_GIT_BASE_SRC"
-    git checkout master || exit 1
-    git pull origin master || exit 1
-}
-
 update_checksum_file()
 {
     if [ "$KERL_BUILD_BACKEND" -eq "git" ];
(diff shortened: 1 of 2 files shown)


kerl
```diff
@@ -782,9 +818,27 @@ do_active()
...
+        tarball_download "$FILENAME.tar.gz"
+    fi
+}
+
+github_download()
+{
     if [ ! -f "$KERL_DOWNLOAD_DIR/$1" ]; then
         echo "Downloading $1 to $KERL_DOWNLOAD_DIR"
-        mkdir -p "$KERL_DOWNLOAD_DIR"
+        curl -L "$OTP_GITHUB_URL/archive/$1" > "$KERL_DOWNLOAD_DIR/$1"
```
mrallen1: Thinking about making this otp_src_$1 too just so its standardized across both download backends

kerl
```diff
@@ -141,6 +160,27 @@ if [ $# -eq 0 ]; then usage; fi
 
 get_releases()
 {
+    if [ "$KERL_BUILD_BACKEND" == "git" ]
```
sanmiguel: Should be
```
if [ "$KERL_BUILD_BACKEND" = "git" ]
```

kerl
```diff
@@ -149,8 +189,13 @@ get_releases()
 
 update_checksum_file()
 {
-    echo "Getting the checksum file from erlang.org..."
-    curl -L $ERLANG_DOWNLOAD_URL/MD5 > "$KERL_DOWNLOAD_DIR"/MD5 || exit 1
+    if [ "$KERL_BUILD_BACKEND" == "git" ];
```
sanmiguel: Likewise `=` not `==`

kerl
```diff
@@ -544,34 +593,31 @@ if [ -n "\$BASH" -o -n "\$ZSH_VERSION" ]; then
...
-        if [ $? -ne 0 ]; then
-            echo "No newest valid release"
-            exit 1
-        fi
-
-        if [ -n "$KERL_INSTALL_MANPAGES" ]; then
-            echo "CAUTION: Fetching and installing newest ($rel) manpages..."
-            do_install_manpages "$rel"
-        fi
+        if [ "$KERL_BUILD_BACKEND" == "tarball" ]; then
```
sanmiguel: `=` again :)

kerl
```diff
@@ -782,9 +828,27 @@ do_active()
 
 download()
 {
+    mkdir -p "$KERL_DOWNLOAD_DIR"
+    if [ "$KERL_BUILD_BACKEND" == "git" ]; then
```
sanmiguel: Last `=` 👍

sanmiguel:
We should change the message during `kerl update releases` that says
```
Run 'kerl update releases' to update this list from erlang.org
```

Maybe just something generic like the following?

```
Run 'kerl update releases' to update this list from your configured provider
```

mrallen1:
Good call out. I'm going to add README updates on this PR too, and test on ubuntu

sanmiguel:
👍

PR #123, opened 2016-04-27
fix crypto start

alexshavelev:
fix for {'EXIT',
                              {undef,
                                  [{crypto,start,[]},
                                   {rebar_core,run,1},
                                   {rebar,main,1},
                                   {escript,run,2},
                                   {escript,start,1},
                                   {init,start_it,1},
                                   {init,start_em,1}]}}


mrallen1:
Thanks for the suggestion. We are pretty wary of adding things like this by default to the main script - but this is a great addition for your own `~/.kerlrc`

PR #125, opened 2016-05-01
Build docs out of tarballs

mrallen1:
Instead of downloading a premade tarball of man pages and HTML docs, this branch will build those artifacts from the source code. This is sort of a companion for #122 - and it's not fully tested yet but it seems to work pretty well on tarballs that have been built out of the Ericsson OTP tags from git hub.

I still need to test the logic for branches that are built off OTP forks (like basho's for instance).


mrallen1:
Oh I got the branches wrong. I'll reopen in a second.

PR #126, opened 2016-05-01
Build and install docs from source

mrallen1:
This code completes the PR in #122 and enables building the man page and HTML artifacts out of a source repository and then installing it where it needs to go.  Still needs testing from a git clone build but works great out of the Ericsson OTP tags.

### DIFF
--- a/kerl
+++ b/kerl
@@ -59,6 +59,9 @@ fi
 if [ -n "$KERL_BUILD_PLT" ]; then
     _KBPLT="$KERL_BUILD_PLT"
 fi
+if [ -n "$KERL_BUILD_DOCS" ]; then
+    _KBD="$KERL_BUILD_DOCS"
+fi
 KERL_CONFIGURE_OPTIONS=
 KERL_CONFIGURE_APPLICATIONS=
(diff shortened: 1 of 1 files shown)


kerl
```diff
@@ -580,34 +593,31 @@ if [ -n "\$BASH" -o -n "\$ZSH_VERSION" ]; then
...
-
-        if [ -n "$KERL_INSTALL_MANPAGES" ]; then
-            echo "CAUTION: Fetching and installing newest ($rel) manpages..."
-            do_install_manpages "$rel"
-        fi
+        if [ "$KERL_BUILD_BACKEND" == "tarball" ]; then
+            if [ "$rel" != "git" ]; then
+                if [ -n "$KERL_INSTALL_MANPAGES" ]; then
+                    echo "Fetching and installing manpages..."
+                    do_install_manpages "$rel"
```
mrallen1: This function name is wrong now. And the one below it.

mrallen1:
OK, I tested it on basho/otp at tag OTP_R16B02_basho10 and everything build 100% successfully, docs and source code.

Issue #113, opened 2016-03-17
kerl can delete home directory


alexshavelev:
I had something like that (kerl was in /home/user/.kerl)

r15b03-1 /home/user/project/
r15b03-1 /home/user/
....

And decided to clean some.
kerl delete installation /home/user/project/ - works good, but
kerl delete installation /home/user/ - delete all data in folder which 'testdisk' can't restore

alexshavelev:
tried again.
it always delete not build but all files with folder. 
after 'kerl delete installation /home/user/project/' folder 'project' doesn't exists

mrallen1:
Sounds like we need to look at the portions that handle removal to be more careful.

alexshavelev:
@mrallen1 i think this bug is also mentioned in https://github.com/yrashk/kerl/issues/52

mrallen1:
I think we just knocked this one out with #127. Please reopen if you find its still an issue. Thanks!

Issue #108, opened 2016-01-23
Errors when installing in a folder with "."


DylanGuedes:
Hello everyone,

After installing erlang with kerl using "kerl install 18.2 ~/.erlang", the installation worked fine but everytime that I interact with erlang I get an error, because of the folder name. The error is:
```
file:path_eval([".","/home/dguedes"],".erlang"): illegal operation on a directory
```
After installing in a folder without dot (".") in its name, everything worked fine. 

cheers.

mrallen1:
This ought to work. We need to figure out what's going wrong and why.

mrallen1:
OK, so after some digging around with @sanmiguel it turns out that `~/.erlang` is a special file that the `erl` command consults on start up. So basically the answer here is unfortunately, *don't do this* - choose a different dot-directory name like `~/.erlangbuilds` or something.  

We are going to add a check into kerl that warns people about this issue because it's somewhat poorly documented or at least not well known. 

For details about the `erl` startup file options, you can read that [here](http://erlang.org/doc/man/erl.html) (See the section about `Configuration`)

DylanGuedes:
Thank you, that's my bad so.

mrallen1:
We just merged #127 which should prevent this accidental mistake in the future. Thanks!

Issue #52, opened 2013-10-15
kerl should not install in non empty directories


peerst:
Currently kerl does a mkdir -p and checks if the target dir is a directory.

It should also check if the directory is empty and refuse to install in nonempty directories.

This avoids messing up and even overwriting directories that e.g. have been accidentially specified.

Since "delete installation" just does a rm -Rf after checking for activate this can get disastrous. E.g. installing by accident in ones home dir, noticing the error and deleting it again => home dir erased

mrallen1:
@sanmiguel This is a great idea

mrallen1:
Knocked this out with #127.

PR #127, opened 2016-05-04
Be more careful about paths before install and delete operations

mrallen1:
As detailed in #113, #108, #52 kerl can make you very very sad if you try to install or remove a build into bad places.  I don't think this is a silver bullet for this problem but it should make it much more difficult to do something *unfortunate* by accident.

Tested on OS X and Ubuntu 14.04 LTS.

### DIFF
--- a/kerl
+++ b/kerl
@@ -864,13 +864,16 @@ is_valid_install_path()
 
 maybe_remove()
 {
-    if [ "$1" = "$HOME" ]; then
+    candidate=$(realpath "$1")
+    canonical_home=$(realpath "$HOME")
+
+    if [ "$candidate" = "$canonical_home" ]; then
         echo "WARNING: You cannot remove an install from $HOME; it's your home directory."
         return 0
     fi
 
     ACTIVE_PATH="$(get_active_path)"
-    if [ "$1" = "$ACTIVE_PATH" ]; then
+    if [ "$candidate" = "$ACTIVE_PATH" ]; then
 	echo "ERROR: You cannot delete the active installation. Deactivate it first."
 	exit 1
     fi
--- a/kerl
+++ b/kerl
@@ -280,7 +280,7 @@ is_valid_installation()
         while read -r l; do
             name=$(echo "$l" | cut -d " " -f 1)
             path=$(echo "$l" | cut -d " " -f 2)
-            if [ "$path" = "$1" ]; then
+            if [ "$name" = "$1" -o "$path" = "$1" ]; then
                 if [ -f "$1"/activate ]; then
                     return 0
                 fi
--- a/kerl
+++ b/kerl
@@ -830,6 +830,7 @@ is_valid_install_path()
 
     candidate=$(realpath "$1")
     canonical_home=$(realpath "$HOME")
+    canonical_base_dir=$(realpath "$KERL_BASE_DIR")
 
     # don't allow installs into home directory
     if [ "$candidate" = "$canonical_home" ]; then
@@ -837,9 +838,9 @@ is_valid_install_path()
         return 1
     fi
 
-    # don't install into .kerl either.
-    if [ "$candidate" = "$canonical_home/.kerl" ]; then
-        echo "ERROR: You cannot install a build into $HOME/.kerl."
+    # don't install into our base directory either.
+    if [ "$candidate" = "$canonical_base_dir" ]; then
+        echo "ERROR: You cannot install a build into $KERL_BASE_DIR."
         return 1
     fi
 
--- a/kerl
+++ b/kerl
@@ -743,24 +743,102 @@ do_deploy()
     echo "kerl_deactivate"
 }
 
+
+# Quoted from https://github.com/mkropat/sh-realpath
+# LICENSE: MIT
+
+realpath() {
+    canonicalize_path "$(resolve_symlinks "$1")"
+}
+
+resolve_symlinks() {
+    _resolve_symlinks "$1"
+}
+
+_resolve_symlinks() {
+    _assert_no_path_cycles "$@" || return
+
+    local dir_context path
+    path=$(readlink -- "$1")
+    if [ $? -eq 0 ]; then
+        dir_context=$(dirname -- "$1")
+        _resolve_symlinks "$(_prepend_dir_context_if_necessary "$dir_context" "$path")" "$@"
+    else
+        printf '%s\n' "$1"
+    fi
+}
+
+_prepend_dir_context_if_necessary() {
+    if [ "$1" = . ]; then
+        printf '%s\n' "$2"
+    else
+        _prepend_path_if_relative "$1" "$2"
+    fi
+}
+
+_prepend_path_if_relative() {
+    case "$2" in
+        /* ) printf '%s\n' "$2" ;;
+         * ) printf '%s\n' "$1/$2" ;;
+    esac
+}
+
+_assert_no_path_cycles() {
+    local target path
+
+    target=$1
+    shift
+
+    for path in "$@"; do
+        if [ "$path" = "$target" ]; then
+            return 1
+        fi
+    done
+}
+
+canonicalize_path() {
+    if [ -d "$1" ]; then
+        _canonicalize_dir_path "$1"
+    else
+        _canonicalize_file_path "$1"
+    fi
+}
+
+_canonicalize_dir_path() {
+    (cd "$1" 2>/dev/null && pwd -P)
+}
+
+_canonicalize_file_path() {
+    local dir file
+    dir=$(dirname -- "$1")
+    file=$(basename -- "$1")
+    (cd "$dir" 2>/dev/null && printf '%s/%s\n' "$(pwd -P)" "$file")
+}
+
+# END QUOTE
+
 is_valid_install_path()
 {
-    # don't allow installs into home directory
-    if [ "$1" = "$HOME" ]; then
-        echo "ERROR: You cannot install a build into $HOME. It's a really bad idea."
-        return 1
-    fi
 
     # don't allow installs into .erlang because
     # it's a special configuration file location
     # for OTP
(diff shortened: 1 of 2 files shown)


kerl
```diff
@@ -733,6 +743,63 @@ do_deploy()
...
+    # don't allow installs into home directory
+    if [ "$1" = "$HOME" ]; then
+        echo "ERROR: You cannot install a build into $HOME. It's a really bad idea."
+        return 1
+    fi
+
+    # don't allow installs into .erlang because
+    # it's a special configuration file location
+    # for OTP
+    if [ "$1" = "$HOME/.erlang" ]; then
```
sanmiguel:
```
$ cd
$ kerl install 18.3 .erlang
Installing Erlang/OTP 18.3 (18.3) in /home/vagrant/.erlang...
```

:-(

Maybe we should pre-process `$1` before calling this function, to make it an absolute path?

sanmiguel: Although having read the `erl` manpage again, perhaps we should just outright ban `.erlang` as an install name:

```
When Erlang/OTP is started, the system searches for a file named .erlang in the directory where Erlang/OTP is started. If not found, the user's home directory  is  searched for an .erlang file.
```

kerl
```diff
@@ -733,6 +743,63 @@ do_deploy()
...
+        echo "ERROR: You cannot install a build into $HOME/.erlang. (It's a special configuration file location for OTP.)"
+        return 1
+    fi
+
+    # don't install into .kerl either.
+    if [ "$1" = "$HOME/.kerl" ]; then
```
sanmiguel: Although the absolute-path-ness assertion applies just as well here.

sanmiguel: Perhaps we can use `readlink -f $1` to compare? Although that might cause issues if `$HOME` lives under a symlink... So we'd probably need to check `if [ "$(readlink -f "$1")" = "$(readlink -f "$HOME/.kerl")" ]`...

Complicated but it should work...

NB We don't want to use `readlink -f ...` to try to sanitise the input beforehand, because it could cause surprises with the resultant install location store in `otp_installations`, I think

kerl
```diff
@@ -276,8 +276,16 @@ get_newest_valid_release()
 
 is_valid_installation()
 {
-    if [ -f "$1"/activate ]; then
-        return 0
+    if [ -f "$KERL_BASE_DIR"/otp_installations ]; then
```
sanmiguel: 👍 💅

kerl
```diff
@@ -733,6 +743,63 @@ do_deploy()
     echo "kerl_deactivate"
 }
 
+is_valid_install_path()
+{
+    # don't allow installs into home directory
+    if [ "$1" = "$HOME" ]; then
```
sanmiguel: Oh, absolute-path-ness thingy here too!

sanmiguel:
Aside from the comments I made, I think this is looking good!

kerl
mrallen1:
```
mallen-rMBP:~ mallen$ readlink -f .
readlink: illegal option -- f
```

Great suggestion. Seems to be the common suggestion from searching but doesn't work on OS X.

mrallen1:
Yeah thinking about using `basename` for some of these issues but I'm not sure how we can exhaustively expand the path into an absolute path. I mean you can subvert this stuff by using something like `./erlang` too.

kerl
sanmiguel: Damn, that felt so right, too. I guess we could go down the rabbit-hole of different things for Dawin and *n*x, but that feels like we're going too far. Do we do it anywhere else?

kerl
sanmiguel: The more I think about it, the more I think banning `.erlang` as a name is reasonable (with an appropriate error message indicating why, of course).

sanmiguel:
The `.erlang` problem can be solved by comparing `$(basename $path)` - the problem is how to avoid the 101 ways to specify `$HOME`.

Although with the change to prevent installing in non-empty directories, you can't install in $HOME anyway... Maybe I'm worrying too much?

mrallen1:
Just comparing `$(basename -- "$path")` to `.erlang` rules out a bunch of legitimate installation locations like `/opt/install/.erlang` (Yes, that's a dumb place to put installs, but it's still fine - OTP won't use that location to find out how to load stuff (right?!)

I found [this](https://github.com/mkropat/sh-realpath) last night and it's awesome... except it won't evaluate paths that don't exist. I'm still honestly not sure how far to push this.

sanmiguel:
Re: `/opt/install/.erlang`: OTP will indeed try to load from that, if you start `erl` from within `/opt/install`... 

```
vagrant@kerl-dev:/opt/install$ pwd
/opt/install
vagrant@kerl-dev:/opt/install$ ls -la
total 12
drwxr-xr-x 3 root root 4096 May 10 16:38 .
drwxr-xr-x 3 root root 4096 May 10 16:38 ..
drwxr-xr-x 2 root root 4096 May 10 16:38 .erlang
vagrant@kerl-dev:/opt/install$ erl
Erlang/OTP 18 [erts-7.3] [source] [64-bit] [smp:4:4] [async-threads:10] [hipe] [kernel-poll:false]


=ERROR REPORT==== 10-May-2016::16:39:28 ===
file:path_eval([".","/home/vagrant"],".erlang"): illegal operation on a directory
Eshell V7.3  (abort with ^G)
1>
```

sanmiguel:
`sh-realpath` looks good in principle I think - we can use `canonicalize_path PATH`. We won't allow installs into a non-empty dir and, for safety, we shouldn't be creating multiple levels of directory right?

But I think we should only use it during checks (on both operands), not to convert paths that e.g. end up in `otp_installations`.

Then we end up with something like:

``` bash
    # don't allow installs into .erlang because
    # it's a special configuration file location
    # for OTP
    if [ "$(basename -- $1)" = ".erlang" ]; then
        echo "ERROR: You cannot install a build into $HOME/.erlang. (It's a special configuration file location for OTP.)"
        return 1
    fi

    # don't install into .kerl either.
    if [ "$(canonicalize_path $1)" = "$(canonicalize_path $HOME/.kerl)" ]; then
        echo "ERROR: You cannot install a build into $HOME/.kerl."
        return 1
    fi
```

Thoughts?

mrallen1:
Yeah I like that... ok - I will work on it tomorrow.

sanmiguel:
:+1:

mrallen1:
@sanmiguel Pretty pleased with this as it stands now.  Please give it a spin.

kerl
```diff
@@ -743,24 +743,102 @@ do_deploy()
...
+
+    # don't allow installs into home directory
+    if [ "$candidate" = "$canonical_home" ]; then
+        echo "ERROR: You cannot install a build into $HOME. It's a really bad idea."
         return 1
     fi
 
     # don't install into .kerl either.
-    if [ "$1" = "$HOME/.kerl" ]; then
+    if [ "$candidate" = "$canonical_home/.kerl" ]; then
```
sanmiguel: Oh! I just realised - this one should be compared against `$KERL_BASE_DIR` right?

sanmiguel:
Looking great! I will give it a whirl today.

kerl
mrallen1: good catch

mrallen1:
Couple more commits wheeeee :)

sanmiguel:
This is looking great!

I have been playing around and have not been able to do shoot myself in the foot yet. :)

I think there is some more work we can do to smooth out the UX a bit (e.g. `kerl delete build ./18.3` doesn't work, you have to do `kerl delete build /full/path/to/18.3`) but this is not the place for that.

mrallen1:
Well, dunno maybe we should fix that now.

sanmiguel:
Ah, found it: L277-291

```
is_valid_installation()
{
    if [ -f "$KERL_BASE_DIR"/otp_installations ]; then
        while read -r l; do
            name=$(echo "$l" | cut -d " " -f 1)
            path=$(echo "$l" | cut -d " " -f 2)
            if [ "$path" = "$1" ]; then
                if [ -f "$1"/activate ]; then
                    return 0
                fi
            fi
        done < "$KERL_BASE_DIR"/otp_installations
    fi
    return 1
}
```

We should be checking against `$name` **and** `$path` I think

sanmiguel:
``` bash
is_valid_installation()
{
    if [ -f "$KERL_BASE_DIR"/otp_installations ]; then
        while read -r l; do
            name=$(echo "$l" | cut -d " " -f 1)
            path=$(echo "$l" | cut -d " " -f 2)
            if [ "$name" = "$1" -o "$path" = "$1" ]; then
                if [ -f "$path"/activate ]; then
                    return 0
                fi
            fi
        done < "$KERL_BASE_DIR"/otp_installations
    fi
    return 1
}
```

sanmiguel:
Something like that?

mrallen1:
Yep.

sanmiguel:
It occurs to me that we still don't cover trying to use a relative path.

How about this:

``` bash
is_valid_installation()
{
    if [ -f "$KERL_BASE_DIR"/otp_installations ]; then
        while read -r l; do
            name=$(echo "$l" | cut -d " " -f 1)
            path=$(echo "$l" | cut -d " " -f 2)
            canonical_path=$(canonicalize_path "$path")
            canonical_install=$(canonicalize_path "$1")
            if [ "$name" = "$1" ] \
                || [ "$path" = "$1" ] \
                || [ "$canonical_path" = "$canonical_install" ] \
                && [ -f "$canonical_install"/activate ]; then
                    return 0
            fi
        done < "$KERL_BASE_DIR"/otp_installations
    fi
    return 1
}
```

mrallen1:
I am not sure what it buys us tbh. A lot of extra complexity and not much benefit IMO.  I am ok with making people exactly match the path name for deletions.

sanmiguel:
Fair enough :)

In which case, I think this is good to go, right?

PR #132, opened 2016-05-13
Remove Lion support code and stop setting CFLAGS

### LINKED ISSUE #131: Version compiled with kerl is more than 50% slower due to CFLAGS
I have noticed that the Erlang compiled via kerl is more than 50% slower than the one compiled manually or installed via homebrew. After some debugging, I found the root cause to be setting CFLAGS:

https://github.com/kerl/kerl/blob/master/kerl#L436-L446

Unless I am misunderstanding the snippet above, it will compile all Erlang versions above R14 with CFLAGS=-O0, which drastically affects the VM performance. I personally don't understand why CFLAGS is being set, specially given the latest Erlang versions work great with GCC LLVM.

josevalim:
If for some reason they are running on Lion, developers can set
CFLAGS on their ~/.kerlrc. Otherwise it is not worth setting
CFLAGS and affect the performance of those running on more
recent OS X versions.

Closes #131

### DIFF
--- a/kerl
+++ b/kerl
@@ -212,19 +212,6 @@ check_releases()
     fi
 }
 
-KERL_NO_LION_SUPPORT="R10B-0 R10B-2 R10B-3 R10B-4 R10B-5 R10B-6 R10B-7
-R10B-8 R10B-9 R11B-0 R11B-1 R11B-2 R11B-3 R11B-4 R11B-5 R12B-0 R12B-1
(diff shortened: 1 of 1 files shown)


mrallen1:
LGTM 👍 

What do you think @sanmiguel

sanmiguel:
Ohai @josevalim !

LGTM :+1:

josevalim:
@sanmiguel hi! that's a well disguised username. :D

Thanks for merging @mrallen1!

mrallen1:
Thanks for the patch! 🍬

PR #133, opened 2016-05-16
Github tags before OTP 17 use underscore

mrallen1:
If the build backend is `git` and you choose a tag that's greater than or equal to 17, then the download tarball is `OTP-XX` otherwise it's `OTP_YY`

Also fixes a problem in the regex of `get_otp_version()` where releases like `R15B03-1` weren't properly handled.

### DIFF
--- a/kerl
+++ b/kerl
@@ -367,7 +367,7 @@ do_git_build()
 
 get_otp_version()
 {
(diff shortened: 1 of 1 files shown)


sanmiguel:
LGTM!

sanmiguel:
👍

PR #136, opened 2016-05-17
Remove path check for deploy

mrallen1:
This is unneeded - but *probably* should be performed on the remote host. Just not sure exactly how to write all of that.

### DIFF
--- a/kerl
+++ b/kerl
@@ -687,9 +687,6 @@ do_deploy()
     host="$1"
 
     assert_valid_installation "$2"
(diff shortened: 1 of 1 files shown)


sanmiguel:
👍

(Apologies for the delay)

Perhaps there's an option of rsync we can utilise to provide similar protection against non-empty dirs etc? It'll be really hard to do the `$HOME` etc checks though, I think.

mrallen1:
Yeah I don't know rsync well enough to say to be honest.

PR #137, opened 2016-05-23
Improve use of 'curl' in kerl script

nalundgaard:
1. __Make kerl use 'curl -o $filename ...' when downloading files__

    Instead of piping output to a file, use curl's '-o' option.

    This change is intended to work around a problem I encountered where I had `-w "\n"` in my `~/.curlrc` file, which caused these commands to produce output that failed the md5 checks.

1. __Use 'curl -q' in 'get_tarball_releases()'__

    This bypasses processing `~/.curlrc`, eliminating the chance that options set in that file may result in errant, unexpected behavior from this curl call.

### DIFF
--- a/kerl
+++ b/kerl
@@ -206,7 +206,7 @@ update_checksum_file()
         return 0
     else
         echo "Getting the checksum file from erlang.org..."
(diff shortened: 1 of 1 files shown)


mrallen1:
:+1: LGTM

sanmiguel:
LGTM 👍

mrallen1:
Thanks for the patch!

Issue #64, opened 2014-03-18
R1404 fails to build on Mac OS X 10.8.4


oubiwann:
OS: Mac OS X 10.8.4
```
$ uname -a
Darwin cahwsx01.local 12.4.0 Darwin Kernel Version 12.4.0:\
 Wed May  1 17:57:12 PDT 2013; \
 root:xnu-2050.24.15~1/RELEASE_X86_64 x86_64
```
I tried both of the following, each of which ended in build failure:
```
$ kerl build R14B04 erlang-r14b04 --disable-hipe
Configured with: --prefix=/Applications/Xcode.app/Contents/Developer/usr \
  --with-gxx-include-dir=/usr/include/c++/4.2.1
Verifying archive checksum...
Checksum verified (4b469729f103f52702bfb1fb24529dc0)
Building Erlang/OTP R14B04 (erlang-r14b04), please wait...
Build failed, see \
  /Users/oubiwann/.kerl/builds/erlang-r14b04/otp_build_R14B04.log
```
and
```
$ KERL_CONFIGURE_OPTIONS="--disable-hipe" kerl build R14B04 erlang-r14b04
Configured with: --prefix=/Applications/Xcode.app/Contents/Developer/usr \
  --with-gxx-include-dir=/usr/include/c++/4.2.1
Verifying archive checksum...
Checksum verified (4b469729f103f52702bfb1fb24529dc0)
Building Erlang/OTP R14B04 (erlang-r14b04), please wait...
Build failed, see \
  /Users/oubiwann/.kerl/builds/erlang-r14b04/otp_build_R14B04.log
```
The final error in the log file is this one:
```
Undefined symbols for architecture i386:
  "_bp_sched2ix", referenced from:
      _load_nif_2 in erl_nif.o
ld: symbol(s) not found for architecture i386
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```
(everything prior to that seems to be warnings).

Note that kerl did actually successfully build R13b04...

mrallen1:
PR #18 addresses this same issue for Solaris. Probably should be extended to handle OS X.

mrallen1:
We need to make sure the patch in #18 is the right one for Darwin but if not we can grab it from erlbrew.

mrallen1:
The patch in #18 does not work on OS X. We will have t grab it from erlbrew.

fabianmartins:
This worked pretty well for me
[...]

PR #138, opened 2016-05-24
Add patches to build older OTP releases

mrallen1:
This PR incorporates code from #18 and addresses #64. The following patches are now inlined into kerl and applied appropriately:

* On OS X since Mavericks (i.e., using clang), when building R14 - fix compiler error
* On Solaris, building R14 - fix networking code
* On any platform with Java 8, building R14 or R15 or R16 - stop linting bad javadoc tags (only applied if kerl is building docs from source)
* On any platform with perl 5.24 building R14 or R15 - fix a fatalized warning while generating beam opcodes

**IMPORTANT**: We do not patch builds from git sources, because there's really no good way to tell if the patches will apply cleanly. (Officially tagged tarballs from OTP github are fine. Some random github repo not so good...)

### DIFF
--- a/kerl
+++ b/kerl
@@ -370,10 +370,9 @@ get_otp_version()
     echo $1 | sed $SED_OPT -e 's/R?([0-9]{1,2})[.AB]?[-.0-9rc]*/\1/'
 }
 
-get_javac_version()
+get_perl_version()
 {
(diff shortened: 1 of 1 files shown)


sanmiguel:
After discussing it I think this LGTM! 👍

PR #140, opened 2016-05-24
Add a version string and command

mrallen1:


sanmiguel:
@mrallen1 Pending your decision either way on #144 , this LGTM and is good to go. 👍

sanmiguel:
Let's not forget to bump `$KERL_VERSION` before we tag after merging. :-)

PR #143, opened 2016-05-29
Print the list of releases/updates with the command column

nihilismus:
Currently:

$ kerl list releases
R10B-0 R10B-10 R10B-1a R10B-2 R10B-3 R10B-4 ...
13B04 R13B R14A R14B01 R14B02 R14B03 R14B04 ...

With these changes:

$ kerl list releases
The available releases are:
R10B-0                                  R14B01
R10B-10                                 R14B02
R10B-1a                                 R14B03
...
R13B                                    18.3
R14A

### DIFF
--- a/kerl
+++ b/kerl
@@ -870,11 +870,15 @@ list_print()
 {
     if [ -f "$KERL_BASE_DIR/otp_$1" ]; then
         if [ "$(wc -l "$KERL_BASE_DIR/otp_$1")" != "0" ]; then
-            if [ -z "$2" ]; then
+            if [ "$2" = "spaces" ]; then
                 cat "$KERL_BASE_DIR/otp_$1"
-            else
-                echo `cat "$KERL_BASE_DIR/otp_$1"`
+                return 0
             fi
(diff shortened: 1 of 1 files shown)


mrallen1:
Thanks for the patch! I like the idea. The default ought to be the "spaces" option, so any automation built around this command won't break.  

So `kerl list releases` should work as it always has, and if you give `kerl list releases columns` it will format the list through the `column` command.

nihilismus:
@mrallen1 I added `kerl list releases columns` and `kerl update releases columns`, what do you think?

mrallen1:
@nihilismus thanks for the patch - I think we're interested in making

`kerl list releases` directly use `column` - we will just drop the column and spaces commands.  That should simplify this patch and make the output look nicer.

mrallen1:
@nihilismus ping - are you interested in making the suggested change?

mrallen1:
Thanks for the contribution, but I'm going to go ahead and close this PR now. Please reopen it if you're still interested.

PR #146, opened 2016-07-12
fix extraction of files for some versions of tar

colrack:
Some versions of tar complain if  `f` parameter is not the last parameter


mrallen1:
LGTM

Thanks!

PR #150, opened 2016-07-18
Fix issue #88

### LINKED ISSUE #88: kerl prompt displays two versions
I've got 17.3 and 17.3.4 installed using kerl. I'm using 17.3, and `kerl prompt` returns

```
 (17.3
17.3.4)
```

I was expecting it to return only ` (17.3)`

mazenharake:
This should fix #88, I noticed though that this also affects the `do_deploy()` which *should* have had a problem but it hasn't been reported anywhere.

I've tested it. 

@rlipscombe / can you try verify?

### DIFF
--- a/kerl
+++ b/kerl
@@ -1009,20 +1009,26 @@ is_valid_install_path()
         return 1
     fi
 
+    INSTALLED_NAME=$(get_name_from_install_path $candidate)
+    if [ -n "$INSTALLED_NAME" ]; then
+        echo "ERROR: Installation ($INSTALLED_NAME) already registered for this location ($1)"
+        return 1
+    fi
+
     # if the install directory exists,
     # do not allow installs into a directory
     # that is not empty
     if [ -e "$1" ]; then
-	if [ -d "$1" ]; then
-	    count=$(ls -la "$1" | wc -l)
-	    if [ $count -ne 3 ]; then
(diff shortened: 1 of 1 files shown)


mazenharake:
Someone also needs to test this on FreeBSD but it "should work"tm

mrallen1:
Need to test this out before I merge it, but it looks good so far.

rlipscombe:
My apologies, I've not had a chance to try this yet. If @mrallen1 is happy with testing it, then I guess we're good. If not, let me know and I'll try to make some time later this week.

mrallen1:
Pathological case here:

```
$ cat ~/.kerl/otp_installations
foo /tmp/test
bar /tmp/test
$ . /tmp/test/activate
$ kerl prompt
(foo
 bar)
```

How do you do this? You install `foo` into `/tmp/test`, then `rm -rf /tmp/test` and then install `bar` into the same location.  The installation check routine needs to make sure the installation location is not already present in our installation file before we allow an install.

mrallen1:
Otherwise, this looks good.

mazenharake:
Added the check to see if the installation path is already registered and also added `-m1` to the grep so that it always returns only 1 row (which *should* always be the only one).

mrallen1:
Thanks! This looks great.

PR #152, opened 2016-07-21
Change path priotiy in activate.fish so that kerl erlang version gets used over the system erlang version

unbalancedparentheses:
The new path should have the kerl binary path first, so that the kerl version has priority over the system erlang version.


rciorba:
LGTM
Thanks!

PR #153, opened 2016-07-22
Fix invalid symlinks when KERL_BUILD_DOCS=true

jhlywa:
The previous code incorrectly copied the generated documentation into `$absdir/lib/erlang/erlang`.  This fix places it in `$absdir/lib/erlang` which is the target of the `man` and `html` symlinks.


mrallen1:
Oh geez. What a think-o. Thanks for the patch!

Issue #154, opened 2016-07-22
KERL_ENABLE_PROMPT to true breaks prompt in zsh


rjcoelho:
I'm using ohmyzsh/prezto and commit 759f73f59879a40279c7b5229e40d21f2c42231e breaks the prompt. It shows buildname twice "(17.5_hipe)(17.5_hipe)"

cat ~/.kerlrc
KERL_ENABLE_PROMPT=true

mrallen1:
Thanks for the report.

mazenharake:
@rjcoelho If you can do this in a terminal and copypaste the results here it would be great:
* Install a build somewhere then immediately activate it press enter then deactivate it.
* Install a build and activate it twice.

rjcoelho:
```
~ ❯❯❯ . /usr/local/opt/kerl/17.5_hipe/activate
(17.5_hipe)(17.5_hipe)kerl_deactivate
~ ❯❯❯ . /usr/local/opt/kerl/17.5_hipe/activate
(17.5_hipe)(17.5_hipe). /usr/local/opt/kerl/17.5_hipe/activate
(17.5_hipe)(17.5_hipe)
```

rjcoelho:
@mazenharake this was working well on previous kerl release. Thanks for the help

mrallen1:
@rjcoelho Did you get a chance to test the fix in #156?

mazenharake:
I downloaded and installed ohmyzsh to verify it. It works.

rjcoelho:
@mazenharake applied 35ea605 and still have the issue using https://github.com/sorin-ionescu/prezto

```
~ ❯❯❯ . /usr/local/opt/kerl/17.5_hipe/activate
(17.5_hipe)(17.5_hipe). /usr/local/opt/kerl/17.5_hipe/activate
(17.5_hipe)(17.5_hipe)
```

This was a regression from previous version (1.3.1) to current (1.3.2)

mazenharake:
Did you do a reinstallation?

rjcoelho:
@mazenharake, you are correct. Patching the activation script solves the issue. Thanks!!

mrallen1:
Great. Thanks for confirming.

PR #156, opened 2016-07-24
Fix issue #154

mazenharake:
Apparently zsh doesn't handle the variable PROMPT very well and overwrites PS1 immediately.

Fixing it by changing the temporary variable name

### DIFF
--- a/kerl
+++ b/kerl
@@ -696,8 +696,8 @@ if [ -n "\$KERL_ENABLE_PROMPT" ]; then
     else
         FRMT="(%BUILDNAME%)"
     fi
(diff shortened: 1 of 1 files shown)


mrallen1:
Thanks for the fix.

rjcoelho:
Applied the patch and still have the issue using https://github.com/sorin-ionescu/prezto

```
~ ❯❯❯ . /usr/local/opt/kerl/17.5_hipe/activate
(17.5_hipe)(17.5_hipe). /usr/local/opt/kerl/17.5_hipe/activate
(17.5_hipe)(17.5_hipe)
```

PR #161, opened 2016-09-25
Ask homebrew for openssl location if available

mrallen1:
If we're building Erlang on { OS X | macOS } of recent vintage, *and* homebrew is installed *and* openssl was installed, ask homebrew where openssl is located and configure OTP to build with that path.

### DIFF
--- a/kerl
+++ b/kerl
@@ -515,18 +515,18 @@ _do_build()
                     echo -n $KERL_CONFIGURE_OPTIONS | grep "ssl" 1>/dev/null 2>&1
                     # Reminder to self: 0 from grep means the string was detected
                     if [ $? -ne 0 ]; then
-                        brew_prefix=$(brew --prefix openssl)
-                        if [ -n "$brew_prefix" -a -d "$brew_prefix" ]; then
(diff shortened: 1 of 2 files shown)


kerl
```diff
@@ -508,16 +508,19 @@ _do_build()
...
             OSVERSION=`uname -r`
             RELVERSION=`get_otp_version "$1"`
             case "$OSVERSION" in
-                15*)
+                16*|15*)
                     echo -n $KERL_CONFIGURE_OPTIONS | grep "ssl" 1>/dev/null 2>&1
                     # Reminder to self: 0 from grep means the string was detected
                     if [ $? -ne 0 ]; then
-                        if [ ! -d /usr/include/openssl -o ! -d /usr/local/include/openssl ]; then
+                        brew_prefix=$(brew --prefix openssl)
```
mrallen1: What if brew doesn't exist? And what if it's not executable?

kerl
```diff
@@ -526,7 +529,7 @@ _do_build()
                                 echo 'WARNING: No OpenSSL library was found in the usual places. Your Erlang will be built without crypto support!'
```
mrallen1: This warning is not useful now that we show the results from `./configure`

mrallen1:
ping @sanmiguel

sanmiguel:
LGTM 👍

Issue #164, opened 2016-10-11
Can't deactivate on stock FreeBSD


essen:
This is probably the same as #93.

```
FreeBSD 10.1-RELEASE-p6 (GENERIC) #0: Tue Feb 24 19:00:21 UTC 2015
$ . ~/erlang/OTP-19.1.3/activate
$ kerl_deactivate
${PATH/...}: Bad substitution
```

"Use bash" is not a valid solution. I will work on a patch when I get a minute.

mrallen1:
Thanks for the PR. Going to close this issue since #165 was merged.

Issue #93, opened 2015-01-24
activete kerl paramete on FreeBSD 10.1


poi119:
I try to use ( .  -> dot command) on ubuntu 12.04 ,it's work fine ,but on target platform on my project is
FreeBSD 10.1 Dot commad(.) didn't work, Pls help me for provide solution on active parameter on kerl on  activate it:
$ . /path/to/install/dir/activate - ubuntu

radekg:
What's your shell? . (source) should work with bash. Are you using sh on FreeBSD 10.1?

poi119:
i used sh shell , i solve bu down grade to freebsd 9.3 and used "su login" then used activate parameter.

radekg:
Just use bash.

poi119:
bash shell ,It's work too, thank for answer!

PR #165, opened 2016-10-11
Use a more portable way to put PATH back to normal

essen:
Fix for #164 (and probably #93).

Works for me on FreeBSD 10.1 and Arch Linux more or less current. No OSX to try right now.

Only `kerl install` is needed to test the changes so it should be quick to test on various other platforms, assuming a build already exists.


mrallen1:
I'll give it a go. Thanks!

mrallen1:
Tested successfully on OS X. Thanks!

essen:
Thanks!

PR #168, opened 2016-12-07
Allow an underscore-prefixed alphanum suffix to the OTP version

sanmiguel:
This helps supporting custom forks of OTP e.g. basho/otp

### DIFF
--- a/kerl
+++ b/kerl
@@ -375,7 +375,7 @@ do_git_build()
 
 get_otp_version()
 {
-    echo $1 | sed $SED_OPT -e 's/R?([0-9]{1,2})[.AB]?[-.0-9rc]*(_[a-zA-Z0-9]*)?/\1/'
+    echo $1 | sed $SED_OPT -e 's/R?([0-9]{1,2}).+/\1/'
 }
 
 get_perl_version()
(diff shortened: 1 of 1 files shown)


sanmiguel:
Fixes this problem:

```
$ kerl build R16B02_basho10 16b02basho10
/usr/local/bin/kerl: line 1209: [: 16_basho10: integer expression expected
Extracting source code
Building Erlang/OTP R16B02_basho10 (16b02basho10), please wait...
/usr/local/bin/kerl: line 469: [: 16_basho10: integer expression expected
```

sanmiguel:
I did consider that, but I was wary of making it too open... More than happy to go with the simpler regex if we can.

BTW - 2-digit? or {1,2}-digit?

mrallen1:
Well, I guess it should be {1,2} but... for all practical purposes we always get 2 digit OTP releases. Kerl won't build R9B anyway (will it?)

sanmiguel:
I don't think I've ever tried anything from before about... R13? 😨

mrallen1:
Think we should go with a simple regex and then merge this :)

sanmiguel:
Regex simplified :)

mrallen1:
ping

Issue #167, opened 2016-11-19
Typo in kerl script


mping:
line https://github.com/kerl/kerl/blob/master/kerl#L1500 has an extraneous '`s`'

`list_remove "$2"s "$escaped"` should read `list_remove "$2" "$escaped"`

mrallen1:
Thanks for the heads up!

mguaypaq:
That `s` is definitely not extraneous! It's because the command is written `kerl delete installation ...` (with singular `installation`) whereas the installations are recorded in the file `$HOME/.kerl/otp_installations` (with plural `installations`).

Now that this change is applied, the command `kerl delete installation ...` will actually delete the installation, but not record it as deleted. As a consequence, future `kerl install ...` commands with the same target directory fail because `kerl` thinks there is already an installation there.

Please revert this commit!

PR #169, opened 2016-12-15
Revert "Remove extraneous 's'"

sanmiguel:
This reverts commit aa237c72060e64a98b0e8a3b3be475387c73f53e
Unfixes the fix for #167


mrallen1:
Yeah you were right I shouldn't have made that original merge so hastily.

Issue #170, opened 2016-12-19
./kerl: line 443: [: : integer expression expected


zfbp:

zfbp:
$ ./kerl build 19.2 erl192
Verifying archive checksum...
Checksum verified (7cdd18a826dd7bda0ca46d1c3b2efca6)
Extracting source code
Building Erlang/OTP 19.2 (erl192), please wait...
./kerl: line 443: [: : integer expression expected
APPLICATIONS DISABLED (See: /var/app/.kerl/builds/erl192/otp_build_19.2.log)
 * jinterface     : No Java compiler found
 * odbc           : ODBC library - link check failed
 * orber          : No C++ compiler found

APPLICATIONS INFORMATION (See: /var/app/.kerl/builds/erl192/otp_build_19.2.log)
 * wx             : wxWidgets not found, wx will NOT be usable

DOCUMENTATION INFORMATION (See: /var/app/.kerl/builds/erl192/otp_build_19.2.log)
 * documentation  :
 *                  xsltproc is missing.
 *                  fop is missing.
 *                  The documentation can not be built.

mrallen1:
Would you please post the output of `perl -v` on your system on this ticket?

mrallen1:
@sanmiguel I think this issue points to the fact that we need to be more careful about assuming that perl and javac exist on a system (and have parseable output) - although Erlang itself will not build if Perl is not installed on a system. (Perl is used to assemble the opcode table for the BEAM emulator.)

zfbp:
```sh
$ perl -v
-bash: perl: command not found

$ cat /etc/centos-release
CentOS release 6.6 (Final)
```

crsuarez:
@mrallen1, I would suggest this change:
https://github.com/kerl/kerl/pull/173

sanmiguel:
Hey there @zfbp !

I'm trying to reproduce this as far as possible, but it's proving difficult: the `configure` script *requires* `perl` in order to build e.g. the opcode table as @mrallen1 said before. When I remove `perl` from my `$PATH` here, I get the following error:

```
$ kerl build 19.2 erl192-2
Verifying archive checksum...
Checksum verified (7cdd18a826dd7bda0ca46d1c3b2efca6)
Extracting source code
Building Erlang/OTP 19.2 (erl192-2), please wait...
/vagrant/kerl/kerl: 383: /vagrant/kerl/kerl: perl: not found
/vagrant/kerl/kerl: 443: [: Illegal number:
[...]

PR #173, opened 2017-01-26
[fix] - get_perl_version improper handling

crsuarez:
Issue:
https://github.com/kerl/kerl/issues/170

### DIFF
--- a/kerl
+++ b/kerl
@@ -380,7 +380,7 @@ get_otp_version()
 
 get_perl_version()
 {
-    perl -v | grep version | sed $SED_OPT -e 's/.*v5\.([0-9]+)\.[0-9].*/\1/'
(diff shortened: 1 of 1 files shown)


kerl
```diff
@@ -380,7 +380,7 @@ get_otp_version()
 
 get_perl_version()
 {
-    perl -v | grep version | sed $SED_OPT -e 's/.*v5\.([0-9]+)\.[0-9].*/\1/'
+    SED_OPT=$(perl -v | grep This\ is | perl -pe 's/.*v5\.([0-9]+)\.[0-9].*/\1/')
```
mrallen1: Thanks for the PR, the `SED_OPT` variable is something we definitely don't want to change its value.

We are going to solve this issue in a different way by asserting perl exists before we execute a build command.

crsuarez: SED_OPT is been set there before, so my change is not changing that behavior. But, I agree.

PR #176, opened 2017-02-03
Assert perl exists before builds (#170)

mrallen1:
Perl 5 is required to compile Erlang from source. It's used (among other things) to assemble BEAM opcodes, NIF jump tables and more. We now assert perl exists on the PATH before we proceed to build Erlang or test for Perl's version string.

We also now test for Java before we try to find its version. Java is not required to compile Erlang though, so we don't output any warnings or errors if its not found. 

This PR addresses #170 and supersedes #173

### DIFF
--- a/kerl
+++ b/kerl
@@ -382,13 +382,35 @@ get_otp_version()
 
 get_perl_version()
 {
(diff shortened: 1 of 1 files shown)


sanmiguel:
I haven't been able to directly test this, but it LGTM.

If you're OK with it as is, let's just merge!

Issue #162, opened 2016-09-30
Put Configure options into a file inside the build


sanmiguel:
When building, we should save the configure options we pass to `./configure` somewhere inside the build so that we can use them later.

This would enable us to:
- 1: add tooling to show you what options an erlang was configured with
- 2: potentially check it automatically when running a build with the same name as an existing one, and flag somehow that the options are different. Related to #158

cuyler:
![giphy](https://cloud.githubusercontent.com/assets/2101159/19090798/f18e86d4-8a34-11e6-8b65-04a7e7600f91.gif)

cuyler:
I got this one.

mrallen1:
Thanks!

mrallen1:
Hey @cuyler ping on this? Have you had a chance to work on this?

mrallen1:
I'm going to work on this issue soon. I know you will soon have your thoughts on way more important things than kerl.

Issue #158, opened 2016-08-12
build uses old config if killed with ctrl-c and restarted


aeyakovenko:
I killed kerl to install some dependencies.  After installing them i still saw this message:

```
$ KERL_CONFIGURE_DISABLE_APPLICATIONS=odbc KERL_CONFIGURE_OPTIONS="--enable-dirty-schedulers --disable-hipe --with-openssl" kerl build 19.0 erlang.19.0.dirty_sched.no_hipe.ssl
Verifying archive checksum...
Checksum verified (f64eefc697d47193444aa224d971f367)
Building Erlang/OTP 19.0 (erlang.19.0.dirty_sched.no_hipe.ssl), please wait...
APPLICATIONS DISABLED (See: /home/ayakovenko/.kerl/builds/erlang.19.0.dirty_sched.no_hipe.ssl/otp_build_19.0.log)
 * crypto         : No usable OpenSSL found
 * jinterface     : No Java compiler found
 * odbc           : ODBC library - link check failed
 * ssh            : No usable OpenSSL found
 * ssl            : No usable OpenSSL found
APPLICATIONS INFORMATION (See: /home/ayakovenko/.kerl/builds/erlang.19.0.dirty_sched.no_hipe.ssl/otp_build_19.0.log)
 * wx             : wxWidgets not found, wx will NOT be usable
DOCUMENTATION INFORMATION (See: /home/ayakovenko/.kerl/builds/erlang.19.0.dirty_sched.no_hipe.ssl/otp_build_19.0.log)
 * documentation  : 
 *                  xsltproc is missing.
 *                  fop is missing.
 *                  The documentation can not be built.
```

after `rm -rf .kerl/builds/erlang.19.0.dirty_sched.no_hipe.ssl` it worked fine

mrallen1:
I see - you let the `configure` script run to completion and then you restarted kerl and it (under the hood) reran `configure` and used its cache'd values for configuration options.

Thanks for the report. I agree that's something we ought to try and improve.

The easiest fix for this would probably be to remove the cache file (if it exists) before `configure` is run.

mrallen1:
This issue should be addressed by #177 so I'm going to close this issue. Please reopen it if there are still problems. Thanks!

PR #177, opened 2017-02-03
Store build configuration in a file

mrallen1:
This change implements storing the build configuration in a file which can later be compared to a potentially new configuration which might trigger the need to re-run the OTP configuration steps.

We also copy the build configuration into the installation directory so that at some future time we can write or enhance a command which shows the build configuration.

Addresses #162 #158

**Note**: Still need to implement any output commands

### DIFF
--- a/kerl
+++ b/kerl
@@ -931,7 +931,7 @@ build_plt()
 
 do_plt()
 {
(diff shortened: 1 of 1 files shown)


sanmiguel:
YES <3

PR #178, opened 2017-02-05
Improve readme

sanmiguel:
- Modernise the examples and sample output
 - Collate all the detail about the available environment variables into a single section, broken down by use (e.g. build options, installation options etc)
 - Add note about installing via homebrew
 - More details about using `KERL_BUILD_BACKEND`


mrallen1:
Oh this is lovely! Thanks!

👍

PR #179, opened 2017-02-05
Fix handling of KERL_INSTALL_HTMLDOCS and KERL_PROMPT

sanmiguel:
- We had no handling for KERL_INSTALL_HTMLDOCS being set in the shell
properly, as KERL_INSTALL_MANPAGES is handled
 - There was a typo in the use of KERL_PROMPT in CSH activate scripts


mrallen1:
👍

PR #183, opened 2017-02-12
Enable travis-CI, add 'kerl path' subcommand

### LINKED ISSUE #182: CI
It would help development, testing & maintenance to have a test suite.
Travis CI has something called a build matrix: it allows starting more than one build per push/tag/pr.

I suggest adding a .travis.yml file that specifies a build matrix for 64bit linux & osx that tests all known OTP versions & some combination of wx, terminfo, openssl versions.
For windows there's appveyor.

The main idea is to have regression testing.

sanmiguel:
Add .travis.yml to resolve #182 

Let's see how this turns out...

Edit to add: this now includes the changes for #190 because I needed them to get Travis-CI happy. As a result, I'm closing #190.

### DIFF
--- a/kerl
+++ b/kerl
@@ -171,7 +171,7 @@ usage()
 {
     echo "kerl: build and install Erlang/OTP"
     echo "usage: $0 <command> [options ...]"
-    echo "\n  <command>       Command to be executed\n\n"
+    printf "\n  <command>       Command to be executed\n\n"
     echo "Valid commands are:"
     echo "  build    Build specified release or git repository"
     echo "  install  Install the specified release at the given location"
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,13 @@ os:
 env:
     global:
         - KERL_BASE_DIR="$TMPDIR"/.kerl
-        - KERL_CONFIGURE_DISABLE_APPLICATIONS="odbc"
+        - KERL_CONFIGURE_DISABLE_APPLICATIONS="odbc wx"
     matrix: 
         - _KERL_VSN=19.2 
+        - _KERL_VSN=18.3
+        - _KERL_VSN=17.5
+        - _KERL_VSN=R16B03-1
+        - _KERL_VSN=R15B03-1
 
 script:
     - ./kerl update releases
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ os:
 env:
   global:
     - KERL_BASE_DIR="$TMPDIR"/.kerl
-    - KERL_CONFIGURE_DISABLE_APPLICATIONS="odbc wx"
+    - KERL_CONFIGURE_DISABLE_APPLICATIONS="odbc"
   matrix: 
     - _KERL_VSN=19.2 
     - _KERL_VSN=18.3
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,5 +14,5 @@ script:
     - travis_wait 30 ./kerl build "$_KERL_VSN" "$_KERL_VSN"
     - ./kerl install "$_KERL_VSN" "install_$_KERL_VSN"
     - ./kerl status
-    - ./kerl delete installation $(kerl path install_$_KERL_VSN)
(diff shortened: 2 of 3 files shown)


sanmiguel:
OK for some reason Travis isn't picking up my ability to enable it for the `kerl` organisation. I've granted access via the settings but still no dice. Since I can enable it for my fork right now, I'm going to continue over there for now.

fenollp:
Hey man! I like it!

We need to find a way to test if the build "links" to openssl correctly (a past issue of Kerl was related to finding openssl).
Is the 4-space legal for yaml? I thought travis/yaml only wanted 2 spaces? well if it works..

```
APPLICATIONS DISABLED (See: /home/travis/.kerl/builds/19.2/otp_build_19.2.log)
 * odbc           : ODBC library - link check failed
APPLICATIONS INFORMATION (See: /home/travis/.kerl/builds/19.2/otp_build_19.2.log)
 * wx             : wxWidgets not found, wx will NOT be usable
```

Maybe we should install these?

sanmiguel:
We definitely should add a step to activate the freshly installed erlang - then we can do start erl and run a crypto-related command that should only pass if openssl was correctly found during installation?

For OSX we may need to add a pre-install step to ensure openssl is installed via brew? Or even a build-matrix axis for brew-installed and stock openssl? Just thinking out loud here...

sanmiguel:
For now I disabled `odbc` for Travis builds - we'll get back to dealing with that later :)

sanmiguel:
Should probably disable `wx` for now too

fenollp:
Agreed to disable `wx`.

Note: `$TRAVIS_OS_NAME` is either `osx` or `linux` as per https://docs.travis-ci.com/user/multi-os/

@mrallen1 any idea which erlang function shows whether OpenSSL is accessible to Erlang?

mrallen1:
@fenollp The only thing I've seen is that if Erlang can't link OpenSSL, it will make starting the crypto application fail at the REPL.

I.e., `1> crypto:start()` dies

sanmiguel:
Note that I've merged the 'kerl path' command from #190 into this branch, as I'm relying on it to test deleting installations

sanmiguel:
OK so I created an old version of kerl (`1.3.2`) locally to create an erlang with broken SSL/`crypto`, and tested `erl -s crypto -s init stop` with that and a working build - provided the activation works as expected during the travis builds, this should be enough to prove that it's able to build against SSL correctly.

I still would like to:
 - [ ] disable `wx`
 - [ ] re-add a few erlang versions to the matrix
 - [ ] add matrix entries for `KERL_BUILD_BACKEND=git` (although maybe we can do this in a separate PR)

sanmiguel:
It looks like there's a problem with building against `ssl` for R15B03-1 - all the other builds passed except this one.

I'm trying a local build to see if I can repro the issue, but I'm tempted to simply cut R15B03-1 from the config for now and fix that separately...

sanmiguel:
Seems like 30 minutes isn't always long enough for an erlang to build on Travis-CI. Perhaps we can bump the wait time somewhat.

mrallen1:
Is this ready for a review now?  Wasn't exactly sure if it's still a WIP.

sanmiguel:
Let me bump the timeout on building to 45 minutes, and I think we're good to go...!

sanmiguel:
@mrallen1 ready for review whenever you are.

kerl
```diff
@@ -171,7 +171,7 @@ usage()
 {
     echo "kerl: build and install Erlang/OTP"
     echo "usage: $0 <command> [options ...]"
-    printf "\n  <command>       Command to be executed\n\n"
```
mrallen1: This does actually change the output. I guess the question is do we care?

```shell
$ printf "\n  <command>       Command to be executed\n\n"

  <command>       Command to be executed

$ echo "\n  <command>       Command to be executed\n\n"
\n  <command>       Command to be executed\n\n
```

sanmiguel: Oh, doh

I swear I tested that, but either way it's not even really related to this PR so I'll roll that out.

Issue #188, opened 2017-02-16
Version number management


sanmiguel:
Folks, we need a better plan for managing which version `kerl` reports. I've just realised we haven't bumped the internal `KERL_VERSION` since `1.3.4`...

I'm not sure what the best way to go about this is, since we just hard-code the version string inside the script anyway. I might be over-thinking it but maybe we can install a hook to bump the version string immediately before tagging?

@mrallen1 thoughts?

mrallen1:
I like the idea of using a hook to automate this but I'm not exactly sure how to make it fool proof.

jadeallenx:
yeah i manually update it - there's some GHA stuff to automate a new tag and stuff but I don't feel like that amount of effort is worth it

PR #189, opened 2017-02-17
Fix string compare, java detection, version string

mrallen1:
Fixing a few issues:

* I forgot to set a new version string on the past two releases. Deeeeerrrpp. (Sorry)
* On macOS when you execute `/usr/bin/java{c}` on a system without Java installed it pops up this super annoying dialog box.  The OTP build does it too, but we can avoid an extra one in kerl by probing `/usr/libexec/java_home` 
* When comparing strings, bash wants the `!=` operator, not the integer comparison `-ne`. Sorry.


sanmiguel:
Sorry for the CI noise - hoping to get back to getting Travis running over the weekend. Right now it will fail because there's no configuration outside of #183 but it's enabled everywhere. Can temporarily disable it if it's getting annoying!

mrallen1:
Ehn, it's fine. Think these would be good to go in sooner than later, so if you get a moment, can you review ?

kerl
```diff
@@ -411,8 +411,10 @@ assert_perl()
 
 get_javac_version()
 {
-    java_loc=$(which javac)
-    if [ -z "$java_loc" ]; then
+    # http://apple.stackexchange.com/a/135079
+    /usr/libexec/java_home >/dev/null 2>&1
```
sanmiguel: Does this hold up on other systems? It doesn't look like it works on my Ubuntu 14.04 server, at least - or is this function only called on MacOS?

kerl
```diff
@@ -589,7 +591,7 @@ _do_build()
     if [ -e "./$KERL_CONFIG_STORAGE_FILENAME.md5" ]; then
         # Compare our current options to the saved ones
         OLD_SUM=$(read -r < "./$KERL_CONFIG_STORAGE_FILENAME.md5")
-        if [ "$SUM" -ne "$OLD_SUM" ]; then
+        if [ "$SUM" != "$OLD_SUM" ]; then
```
sanmiguel: Good catch!

kerl
```diff
@@ -24,7 +24,7 @@
 
 unset ERL_TOP
 
-KERL_VERSION="1.3.4"
+KERL_VERSION="1.4.1"
```
sanmiguel: Have you had any thoughts on #188 ?

kerl
mrallen1: Does that file not exist when you install OpenJDK on a Linux box? Hmm. Maybe for Linux we just directly check to see if javac exists...

kerl
sanmiguel: Also let's not forget to bump this to `1.4.2` before creating the next release :)

kerl
sanmiguel: Can we roll this java-related change into a separate PR? Then we can go ahead and merge the rest of this and figure out the java stuff without blocking it?

mrallen1: Yeah, good idea.

kerl
mrallen1: 👍

mrallen1:
Closing in favor of #191

PR #191, opened 2017-02-21
Use string not-equal instead of integer not-equal

mrallen1:
Salvaging the only necessary fix from #189


sanmiguel:
Looks like we're just waiting for Travis-CI to serve up some MacOS hardware for us

PR #196, opened 2017-03-24
Use xargs to set env vars en masse

### LINKED ISSUE #194: Replace env var crunching with neat xargs trick
Use the trick here: https://gist.github.com/judy2k/7656bfe3b322d669ef75364a46327836

to be able to avoid having to manually touch every single env var in https://github.com/kerl/kerl/blob/master/kerl#L51-L138

sanmiguel:
Fixes #194 

/cc @mrallen1

### DIFF
--- a/kerl
+++ b/kerl
@@ -24,118 +24,40 @@
 
 unset ERL_TOP
 
-KERL_VERSION="1.4.2"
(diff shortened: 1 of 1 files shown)


sanmiguel:
Withdrawing. This needs some more thought.

Issue #197, opened 2017-03-28
Cannot build R15B03-1 on Ubuntu16.10 64bit


fenollp:
Build log

```
kerl build R15B03-1 15.3.1
Downloading otp_src_R15B03-1.tar.gz to /home/pete/.kerl/archives
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   178  100   178    0     0    205      0 --:--:-- --:--:-- --:--:--   205
100 72.5M  100 72.5M    0     0   347k      0  0:03:33  0:03:33 --:--:--  376k
Getting checksum file from erlang.org...
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   178  100   178    0     0    290      0 --:--:-- --:--:-- --:--:--   290
100 26653  100 26653    0     0  18060      0  0:00:01  0:00:01 --:--:-- 11.7M
Verifying archive checksum...
Checksum verified (eccd1e6dda6132993555e088005019f2)
Extracting source code
Building Erlang/OTP R15B03-1 (15.3.1), please wait...
Build failed.
	-emulator /home/pete/.kerl/builds/15.3.1/otp_src_R15B03-1/lib/compiler/src/genop.tab beam/ops.tab hipe/hipe_ops.tab && echo /home/pete/.kerl/builds/15.3.1/otp_src_R15B03-1/lib/compiler/src/genop.tab beam/ops.tab hipe/hipe_ops.tab utils/beam_makeops >x86_64-unknown-linux-gnu/opt/plain/OPCODES-GENERATED
Can't use 'defined(@array)' (Maybe you should just omit the defined()?) at utils/beam_makeops line 1714.
x86_64-unknown-linux-gnu/Makefile:501: recipe for target 'x86_64-unknown-linux-gnu/opt/plain/OPCODES-GENERATED' failed
make[2]: *** [x86_64-unknown-linux-gnu/opt/plain/OPCODES-GENERATED] Error 255
make[2]: Leaving directory '/home/pete/.kerl/builds/15.3.1/otp_src_R15B03-1/erts/emulator'
/home/pete/.kerl/builds/15.3.1/otp_src_R15B03-1/make/run_make.mk:40: recipe for target 'generate' failed
make[1]: *** [generate] Error 2
[...]

PR #198, opened 2017-03-30
Apply makeop patch for perls >= 5.22

mrallen1:
Addresses #197 

Perls >= 5.22 need to have the makeop patch applied when they are building older OTPs.


fenollp:
Trying this right now on R15B03-1.
Shouldn't that version be on Travis too?

mrallen1:
Probably should be.

PR #200, opened 2017-04-13
fix bad usage of 'read' command

vladdu:
'read' never returns a value as a result, but through variables specified as arguments


mrallen1:
Thanks for the fix!

PR #203, opened 2017-05-10
Fix perl version retrieval

jamhed:
On Centos 6 `perl -v` returns `This is perl, v5.10.1` and this
string fails to be parsed by kerl. But `perl -V` returns expected format
on Centos 6, and also for modern versions of Perl.

### DIFF
--- a/kerl
+++ b/kerl
@@ -391,7 +391,7 @@ get_otp_version()
 get_perl_version()
 {
     if assert_perl; then
-        perl -v | grep version | sed $SED_OPT -e 's/.*v5\.([0-9]+)\.[0-9].*/\1/'
+        perl -V | grep version | sed $SED_OPT -e 's/.*v5\.([0-9]+)\.[0-9].*/\1/'
     else
         echo "FATAL: I couldn't find perl which is required to compile Erlang."
         exit 1


sanmiguel:
Unfortunately `perl -V` returns a totally differently formatted blob of text on OSX (Yosemite, at least - don't have Sierra to hand right now), which doesn't even contain the `v5.mm.pp` format version string, but does contain the string `version` a bunch. Unfortunately we actually don't bail if there's an error here, so while all the CI tests passed, they actually all printed errors for this on Ubuntu and OSX.

I think we either need to re-examine what we're really checking for, or make `get_perl_version` more aware of OS etc. @mrallen1 thoughts?

jamhed:
@sanmiguel, current kerl expects `version\s(\d+)` format, so it should work on OSX too.

mrallen1:
The middle number is only thing that we need to know, so the regex is probably *too* complicated and should be simplified. 

I don't want to parse through all of the `-V` output since it's somewhat platform dependent unfortunately.  Every Linux vendor seems have patches for munging its output in different ways.

mrallen1:
Closing in favor of #204.  Thank you for the patch and for bringing this issue up!

PR #204, opened 2017-05-23
Get perl version in a more portable way

mrallen1:
This is a more portable way than regex to extract a perl minor version release number. 

Intended to address #203.

P.S. I feel shame for this "fix" but it works great.


sanmiguel:
I approve this abomination.

👍

PR #206, opened 2017-05-23
Conditionally switch where docs symlink to

### LINKED ISSUE #193: Incorrect docs symlinks in some cases
In at least some cases, the symlinks we create to `html` and `man` dirs are incorrect:

```
./html -> /Users/sanmiguel/kerl/19.2.3/lib/erlang/doc
./man -> /Users/sanmiguel/kerl/19.2.3/lib/erlang/man
```

These should be

```
./html -> /Users/sanmiguel/kerl/19.2.3/lib/doc
./man -> /Users/sanmiguel/kerl/19.2.3/lib/man
```

sanmiguel:
In some cases the docs are built without the 'erlang' subdir.
Fixes #193

### DIFF
--- a/kerl
+++ b/kerl
@@ -889,8 +889,13 @@ ACTIVATE_CSH
         if [ -d "$DOC_DIR" ]; then
             echo "Installing docs..."
             cp $CP_OPT "$DOC_DIR/" "$absdir/lib"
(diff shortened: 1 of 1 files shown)


mrallen1:
LGTM. I think this is the right way to solve this problem even though you said it was "the easy way."

PR #214, opened 2017-06-22
Add automatic shell history environment flag for OTP >= 20

mrallen1:
Based on a [tweet][1] from @ferd, add automatic shell history for OTP >= 20. Only works for bash/zsh currently. Probably want fish shell support and t/csh support too?  Or maybe we should think about a different way to include this feature.

[1]: https://twitter.com/mononcqc/status/877544929496629248

### DIFF
--- a/kerl
+++ b/kerl
@@ -768,6 +768,12 @@ REBAR_PLT_DIR="$absdir"
 export REBAR_PLT_DIR
 _KERL_ACTIVE_DIR="$absdir"
 export _KERL_ACTIVE_DIR
+# https://twitter.com/mononcqc/status/877544929496629248
+export _KERL_SAVED_ERL_AFLAGS=" \$ERL_AFLAGS"
+kernel_history=\$(echo "\$ERL_AFLAGS" | grep "kernel shell_history")
+if [ -z "\$kernel_history" ]; then
+    export ERL_AFLAGS="-kernel shell_history enabled \$ERL_AFLAGS"
+fi
 if [ -f "$KERL_CONFIG" ]; then . "$KERL_CONFIG"; fi
 if [ -n "\$KERL_ENABLE_PROMPT" ]; then
     _KERL_SAVED_PS1="\$PS1"
@@ -899,19 +905,6 @@ endif
 
 rehash
 ACTIVATE_CSH
-
-    otpver=$(get_otp_version "$rel")
-    if [ $otpver -ge 20 ]; then
-        # https://twitter.com/mononcqc/status/877544929496629248
-        cat << OTP_20_ADDITION >> "$absdir/activate"
-_KERL_SAVED_ERL_AFLAGS=" \$ERL_AFLAGS"
-export _KERL_SAVED_ERL_AFLAGS
-kernel_history=\$(echo "\$ERL_AFLAGS" | grep "kernel shell_history")
-if [ -z "\$kernel_history" ]; then
-    export ERL_AFLAGS="-kernel shell_history enabled \$ERL_AFLAGS"
-fi
-OTP_20_ADDITION
-    fi
 
     if [ -n "$KERL_BUILD_DOCS" ]; then
         DOC_DIR="$KERL_BUILD_DIR/$1/release_$rel/lib/erlang"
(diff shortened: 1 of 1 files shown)


mrallen1:
Tagging @rciorba for possible fish support

mrallen1:
Tagging @ZeWaren for possible t/csh support

kerl
```diff
@@ -894,6 +899,17 @@ endif
...
 rehash
 ACTIVATE_CSH
+
+    otpver=$(get_otp_version "$rel")
+    if [ $otpver -ge 20 ]; then
+        # https://twitter.com/mononcqc/status/877544929496629248
+        cat << OTP_20_ADDITION >> "$absdir/activate"
+_KERL_SAVED_ERL_AFLAGS="\$ERL_AFLAGS"
+export _KERL_SAVED_ERL_AFLAGS
+export ERL_AFLAGS="-kernel shell_history_enabled \$ERL_AFLAGS"
```
ferd: `-kernel shell_history enabled`

mrallen1: Awesome. Thanks!

mrallen1:
The deactivation isn't quite right yet. Experimenting.

kerl
```diff
@@ -894,6 +899,17 @@ endif
 
 rehash
 ACTIVATE_CSH
+
+    otpver=$(get_otp_version "$rel")
+    if [ $otpver -ge 20 ]; then
+        # https://twitter.com/mononcqc/status/877544929496629248
+        cat << OTP_20_ADDITION >> "$absdir/activate"
+_KERL_SAVED_ERL_AFLAGS="\$ERL_AFLAGS"
```
sanmiguel: The issue you saw with deactivation not working properly comes from 2 parts:
 - if `$ERL_AFLAGS` is empty, `$_KERL_SAVED_ERL_AFLAGS` will be empty, the `[ -n "$_KERL_SAVED_ERL_AFLAGS" ]` check will fail, and `ERL_AFLAGS` will not be reset
 - we do not check that `$ERL_AFLAGS` does not already contain `-kernel shell_history` somewhere

kerl
```diff
@@ -894,6 +899,17 @@ endif
...
+export _KERL_SAVED_ERL_AFLAGS
+export ERL_AFLAGS="-kernel shell_history enabled \$ERL_AFLAGS"
```
sanmiguel: We should wrap this with a check for `-kernel shell_history` not already being present. Something along the lines of the following I guess:

```
if [[ "\$ERL_AFLAGS" != *-kernel\ shell_history* ]]; then
    export ERL_AFLAGS="-kernel shell_history enabled \$ERL_AFLAGS"
fi
```

Need to verify that glob match is portable enough, though.

kerl
sanmiguel:
```
_KERL_SAVED_ERL_AFLAGS=" \$ERL_AFLAGS"
```

kerl
```diff
@@ -894,6 +899,17 @@ endif
 
 rehash
 ACTIVATE_CSH
+
+    otpver=$(get_otp_version "$rel")
```
sanmiguel: If this kernel flag has no effect on OTPs before 20, I'd be inclined to do away with the conditional on `get_otp_version` and just always dump this in `activate` etc. I think it might keep things a bunch neater. What do you reckon @mrallen1 ?

kerl
sanmiguel: It's dirty, but it works!™

kerl
mrallen1: I agree. If it's ok to set this all the time, we should just set it all the time (if it's not already set.)

kerl
mrallen1: 👍

kerl
mrallen1: 👍

rciorba:
Sorry for the silence, the notification got lost amongst all the 9-5 ones. I'll add the fish support

mrallen1:
I'll try to do some work on this tonight.

mrallen1:
@sanmiguel I know you're probably jetlagged af but when you get a chance, take a look.

PR #219, opened 2017-07-31
Fix "kerl plt" to show actual path

shino:
Without the fix, `kerl status` shows proper plt path but `kerl plt` does not.


mrallen1:
So happy to see @shino pop up in a PR again! 👋 Thank you for the fix!

Issue #221, opened 2017-08-02
Test Linux systems for required packages before attempting builds


mrallen1:
Most of our support tickets on Linux come from builds that fail because one or more of the required debs or RPMs have not been installed before attempting a build.

Kerl ought to check the package management system to see if the required packages have been installed before attempting to build Erlang on some Linux platforms.

mrallen1:
I do not want to support **every** Linux distribution in existence but some set of {Fedora, CentOS, RHEL, Debian, Ubuntu} we could test to see if the package manager contains the development libraries Erlang expects to use/find before we attempt a build.

mrallen1:
Or maybe instead of trying to figure out what Linux we're running on, it just checks to see if the distribution uses `rpm` or `dpkg` and then checks if the necessary packages exist.

mrallen1:
Going to close this ticket since we merged #222

PR #222, opened 2017-08-03
Attempt to detect when prerequisite packages are not on a Linux

mrallen1:
Addresses #221 

This is a best effort attempt - it doesn't even fail to attempt a build. It just outputs a warning and continues on its merry way.  The goal is that if a build fails, it will be more obvious why.


mrallen1:
Hey! If you're [here because of my tweet](https://twitter.com/bytemeorg/status/894381108770549760) it would be super awesome if you could test it on a system that's "fresh" - not used for development before. Just want to confirm that it detects the absence of the needed packages correctly and attempts to build Erlang anyway for the most part.

Your thoughts on the behavior would be appreciated too. I'm biased toward people > tools but sometimes making a process broken until a prereq is satisfied has a certain logic to it too.

PR #225, opened 2017-08-11
suggest the correct activate script based on parent process command

### LINKED ISSUE #224: After installation, kerl points to sh script in Fish shell
When kerl installs an Erlang version in Fish shell, it suggests executing `. <install_path>/activate`, while it should suggest executing `. <install_path>/activate.fish`.

rciorba:
Looks at parent process command and suggests sourcing the appropriate activate script for fish and csh.

fixes #224

### DIFF
--- a/kerl
+++ b/kerl
@@ -933,8 +933,8 @@ ACTIVATE_CSH
     fi
 
     PID=$$
-    PPID=$(ps -p $PID -o ppid | grep -E "[0-9]+" -o)
-    PARENT_CMD=$(ps -p $PPID -o comm | tail -n 1)
+    PARENT_PID=$(ps -p $PID -o ppid | grep -E "[0-9]+" -o)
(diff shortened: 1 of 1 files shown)


kerl
```diff
@@ -932,8 +932,23 @@ ACTIVATE_CSH
         build_plt "$absdir"
     fi
 
+    PID=$$
+    PPID=$(ps -p $PID -o ppid | grep -E "[0-9]+" -o)
```
mrallen1: Clever I like this. Just wonder if it's portable across Mac and Linux.  Might need to stuff those flags up with the SED_OPTS and MD5_OPTS above if TravisCI complains.

rciorba: I've tested them on freebsd and linux. Fingers crossed for Mac :)

mrallen1:
Thanks for the fix on MacOS @rciorba! 

@sanmiguel could you run some basic smoke testing on this branch and merge it if it looks good?  Thanks!

PR #228, opened 2017-09-16
Link to full list of configuration options.

mpmiszczyk:


mrallen1:
Thank you for these corrections and additions!

PR #230, opened 2017-09-27
DED_LDFLAGS tomfoolery for macOS 10.13

joedevivo:
My old builds on an upgraded laptop were giving me an error like the one [here](https://bugs.erlang.org/browse/ERL-439). They fixed it in erlang/otp#1501, but even backported to maintenance branches, any versioned source tarball won't be able to build SSL because the `DED_LDFLAGS` are no longer compatible.

I haven't gone back far enough to check, but this definitely works on OTP 20.1. I'll test it on more builds throughout the day, but I thought I'd get it into your hands sooner than later.

### DIFF
--- a/kerl
+++ b/kerl
@@ -538,6 +538,16 @@ do_normal_build()
     list_add builds "$1,$2"
 }
 
+_flags(){
+  host=$(./erts/autoconf/config.guess)
+  DARWIN_VERSION=$(uname -r | sed $SED_OPT -e 's/R?([0-9]{1,2}).+/\1/')
+  if [ "$DARWIN_VERSION" -ge 17 ]; then
+    CFLAGS="$CFLAGS" DED_LD="$CC" DED_LDFLAGS="-m64 -bundle -bundle_loader ${ERL_TOP}/bin/$host/beam.smp" $@
+  else
+    CFLAGS="$CFLAGS" $@
+  fi
+}
+
 _do_build()
 {
     case "$KERL_SYSTEM" in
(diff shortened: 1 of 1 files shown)


mrallen1:
Thanks for the patch! Definitely a lot of good ideas in here. I will take a closer look at this tomorrow!

kpy3:
It seems this patch doesn't work against Erlang 17 (tried it against 17.5.6.9)

joedevivo:
I don't know how it was working before, but after a reboot it wasn't even working for 20.1. It turns out that `kerl` has no access to the `$CC` environment variable that autoconf chooses.

This is only a big deal because otp's `erts/configure.in` file only lets you override `DED_LDFLAGS` if you also override `DED_LD`, so we have to set it to something, and that something needs to be whatever `CC` is. I've fixed this temporarily by hard coding both to `clang` for High Sierra only. There's got to be a better way, but I'm now way out of my depth with autoconf.

joedevivo:
@kpy3 I was able to build 17.5.6.9 after this last commit.

mrallen1:
This whole thread is making me 😢 Why can't we have nice things 🍎 ???!!!

kpy3:
@joedevivo tried updated PR (with https://github.com/kerl/kerl/pull/232) and it seems to work now. The only strange thing I have to notice is 

```
===> Error in dialyzing apps: Could not get abstract code for file: /Users/yes/.kerl/install/17.5/lib/crypto-3.5/ebin/crypto.beam (please recompile it with +debug_info)
```

My build flags are: `KERL_CONFIGURE_OPTIONS="--enable-smp-support -with-ssl=`brew --prefix openssl` --enable-threads --enable-kernel-poll --with-dynamic-trace=dtrace --enable-dirty-schedulers"`

Any thoughts on this?

mrallen1:
Thanks for the patch, I am closing this in favor of #234

PR #232, opened 2017-09-28
Fix point arithmetic error in wx_impl.cpp for Erlang with XCode CLT 9.0

kpy3:
Compilation of Erlang with XCode CLT 9.0 fails with bad point arithmetic error. This patch fixes that situation.

### DIFF
--- a/kerl
+++ b/kerl
@@ -482,6 +482,12 @@ maybe_patch_all()
             15)
                 apply_r15_beam_makeops_patch >> "$LOGFILE"
                 ;;
+            16)
+                apply_r16_wx_ptr_patch >> "$LOGFILE"
+                ;;
+            17|18|19)
+                apply_wx_ptr_patch >> "$LOGFILE"
+                ;;
             *)
                 ;;
         esac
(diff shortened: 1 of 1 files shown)


mrallen1:
Thanks for the patch. I guess this is already fixed upstream in > 18?

kpy3:
Ouch... it fixed in 20.0... :( will update this PR...

mrallen1:
Thank you!!

kpy3:
You are welcome!

nalundgaard:
@kpy3 I pulled this down today on my MacOS 10.13 High Sierra installation to try out with the latest erlang 19 (19.3.6.2). I pulled it down into a local branch with #230 merged in as well, and I see this:

```
$ kerl build 19.3.6.2 19.3.6.2
Extracting source code
Building Erlang/OTP 19.3.6.2 (19.3.6.2), please wait...
APPLICATIONS DISABLED (See: /Users/nal/.kerl/builds/19.3.6.2/otp_build_19.3.6.2.log)
 * odbc           : ODBC library - header check failed

DOCUMENTATION INFORMATION (See: /Users/nal/.kerl/builds/19.3.6.2/otp_build_19.3.6.2.log)
 * documentation  :
 *                  fop is missing.
 *                  Using fakefop to generate placeholder PDF files.

Build failed.
  if((index < memenv->next) && ((index == 0) || (temp > NULL)))
                                                 ~~~~ ^ ~~~~
wxe_impl.cpp:681:55: error: ordered comparison between pointer and zero ('void *' and 'long')
  if((index < memenv->next) && ((index == 0) || (temp > NULL))) {
                                                 ~~~~ ^ ~~~~
2 errors generated.
make[3]: *** [x86_64-apple-darwin17.0.0/wxe_impl.o] Error 1
make[2]: *** [opt] Error 2
make[1]: *** [opt] Error 2
make: *** [libs] Error 2

Please see /Users/nal/.kerl/builds/19.3.6.2/otp_build_19.3.6.2.log for full details.
```

Maybe there is an issue with the logic for triggering the patch?

nalundgaard:
Ah, I figured why this doesn't work for me:

```sh
 perlver=$(get_perl_version)
     if [ "$perlver" -ge 22 ]; then
```

This ends up executing `perl -e 'print int(($] - 5)*1000)'`, which yields this on macOS 10.13 (at least, mine):

```
$ perl -e 'print int(($] - 5)*1000)'
18
$ which perl
/usr/bin/perl
$ perl --version

This is perl 5, version 18, subversion 2 (v5.18.2) built for darwin-thread-multi-2level
(with 2 registered patches, see perl -V for more detail)

Copyright 1987-2013, Larry Wall

Perl may be copied only under the terms of either the Artistic License or the
GNU General Public License, which may be found in the Perl 5 source kit.

Complete documentation for Perl, including FAQ lists, should be found on
this system using "man perl" or "perldoc perl".  If you have access to the
Internet, point your browser at http://www.perl.org/, the Perl Home Page.
```

@mrallen1 I guess I need to brew install perl for all this magic?

kpy3:
Actually I don't know about this limitation much, I use perl from brew

```
% perl -e 'print int(($] - 5)*1000)'
25
```

and it works just fine.

nalundgaard:
Alright, I RTFupdatedM and saw that I do need a newer perl. I did that and I've verified that this patch solves the issue for R16B03-1, 19.3.6.2, 18.3.4.5, 17.5.6.9... 👍 from me, I think this should be merged.

mrallen1:
I appreciate the patches, but the suggestion for when and how to trigger them is not where I want them to land.

These patches are totally orthogonal to what version of Perl might or might not be installed on a given system.  It looks like these patches only apply to Darwin builds too, so at a minimum, they ought to be considered only when we're building on Darwin and only when we're using clang 9.

mrallen1:
So tl;dr I think we do want to include them, but they need to be considered at a different point in the build process.

kpy3:
Perl in High Sierra is installed by default, but a bit outdated, so `kerl` should warn (on even fail) in that situation somewhere in `get_perl_version` or (better) in `maybe_patch_darwin` when building Erlang <= 19 on High Sierra.

mrallen1:
OK, thanks for the patches - I am closing this in favor of #235

PR #241, opened 2017-11-08
Fix requirements check on Debian-based systems

dmitrivereshchagin:
Underscore was missing in function name.


mrallen1:
Wow, an embarrassing error on my part.  Thank you for fixing it!!

PR #242, opened 2017-11-08
Update README

mrallen1:
Clean up the README in a few areas:

* Fix verb tense
* Revise OTP support policy to make it clear we're supporting R15
* Add some text about building crypto on Red Hat

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,10 +10,10 @@ installation is as fast as possible.
 
 OTP Support Policy
 ------------------
-As of 2017 May 24, we are supporting OTP back to R14B04. Older builds
+As of 2017 November 8, we are supporting OTP builds back to R15. Older builds
 may or may not work. We will advance release support as new releases of OTP
-become available.  For example, when OTP 20 is released, we will support Erlang
-builds R15 and newer.
+become available.  For example, when OTP 21 is released, we will support Erlang
+builds R16 and newer.
 
 Triage cadence
 --------------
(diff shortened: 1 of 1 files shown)


fenollp:
If R15 is supported how about adding it to Travis matrix?

mrallen1:
Our travis jobs are already running for hours and hours before they complete.  If anything I'd like to streamline them quite a bit.  Adding another OTP version will require another hour-plus worth of run time before the test completes.

mrallen1:
Our travis jobs are already running for hours and hours before they complete.  If anything I'd like to streamline them quite a bit.  Adding another OTP version will require another hour-plus worth of run time before the test completes.

mrallen1:
Our travis jobs are already running for hours and hours before they complete.  If anything I'd like to streamline them quite a bit.  Adding another OTP version will require another hour-plus worth of run time before the test completes.

mrallen1:
Our travis jobs are already running for hours and hours before they complete.  If anything I'd like to streamline them quite a bit.  Adding another OTP version will require another hour-plus worth of run time before the test completes.

fenollp:
Let's add more CIs then.
Here's a few of them https://github.com/ligurio/Continuous-Integration-services/blob/master/continuous-integration-services-list.md

mrallen1:
I might welcome some alternatives but I don't have the time to investigate a bunch of them, sorry.

PR #243, opened 2017-11-08
Shellcheck

fenollp:
WIP: Travis would always fail on my repo, maybe on yours it passes?
Just added the badge BTW.

This PR
1. adds a `shellcheck` pass on the `kerl` script, to identify Shell issues
1. adds testing of 20.1 (OTP 20 wasn't tested yet)
1. shellcheck is run only on one of the jobs of the build matrix
1. `set -e` is added to better & faster detect script errors

I don't think this PR will have the fix for the shellcheck warnings, but it's a good start just to have them.
Once I'm finished with this PR I'll probably replace the shellcheck command with `shellcheck ./kerl || true`.

Tell me what you think :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,13 @@
+language: bash
+sudo: false
+
+addons:
+  apt:
+    sources:
+      - debian-sid # for: shellCheck
+    packages:
+      - shellcheck
+
 os:
   - linux
   - osx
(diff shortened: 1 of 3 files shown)


mrallen1:
Thanks for this PR.  I am a bit neutral about shellcheck to be honest. I often find it a bit too pedantic. Genuinely curious what value does it bring for you?

mrallen1:
Thanks for this PR.  I am a bit neutral about shellcheck to be honest. I often find it a bit too pedantic. Genuinely curious what value does it bring for you?

fenollp:
I understand it can feel too pedantic, but considering that not all possible deployment targets can be tested I feel that fixing most of the issues it finds comforting. I can fix them given some time (and an already passing CI). 

Some things can feel pedantic "error code of `cd` was not checked" or the sometimes false positive "this was not in double quotes" but most of the time it doesn't cost much to fix these.

I created this PR after seeing https://github.com/kerl/kerl/pull/241 the kind of thing shellcheck would catch.
The first thing it caught for example is interesting already:

```
In ./kerl line 30:
GREP_OPTIONS=''
^-- SC2034: GREP_OPTIONS appears unused. Verify it or export it.
```
this really cannot be used by any sub command unless exported.

fenollp:
Ready :)

fenollp:
@mrallen1 News on this?

mrallen1:
My biggest concern with the patch is that I don't normally run shellcheck as part of my release process, so I'm just worried that I'm signing up for a ton of test failures.

fenollp:
Oh you have additional tests that you run before release that are not run by CI?
I'm interested :)

PR #244, opened 2017-11-12
Fix deleting installation by name

essen:
`kerl delete installation` allows deleting by name or path.
Problem is, if you provide the name it tries to check for a
file `"$name"/activate` instead of using the path, and fails.


essen:
Seems there's also an issue when doing the actual deletion.

essen:
OK this has also been fixed. Deleting installation by name now works for me.

essen:
I notice that the message says "The installation in ...", that part is still incorrect.

mrallen1:
Thanks for the patch! Testing...

Issue #237, opened 2017-10-23
Kerl silently creates a build without crypto when no libssl headers are availible


kanes115:
Kerl still does not ask a user if he really wants a build without crypto included (when he lacks libssl headers). Such a build is pretty useless as e.g. rebar3 won't work without crypto.app. That would be great to give a warning in such situation.

I came across this problem on Ubuntu 16.04.

mrallen1:
We have been outputting the configure analysis of build options since July 2016 which pretty clearly shows whether crypto will be built or not.  We also have a pending PR #221 to check Linux systems for required libraries/packages before a build starts.

What sort of warning do you think would be suitable? What additional information would you like kerl to warn you about?

mrallen1:
Also curious if you're running into this [issue][1]

[1]: https://github.com/kerl/kerl/issues/226#issuecomment-327915702

mrallen1:
I think we've addressed this issue about as well we can, so I'm going to close this issue. Please open a new issue if you're still having trouble. Thanks.

Issue #226, opened 2017-08-28
Erlang doesn't build correctly for certain versions of OpenSSL 1.1.x


fenollp:
debian: netinstall cd image 9.1.0 
openssl: 1.1.0f-3 amd64
erlang: 18.2 compiled & installed with latest Kerl

The error: crypto app is not found.
Do we need to install openssl-dev or some other thing?

Tell me if you need more info.

mrallen1:
Yes. I believe the library is called libssl-dev. Would you try the code on PR #222? Curious how it works out for you.

fenollp:
@claytonfinney plz try this ^

mpmiszczyk:
I believe you need to downgrade your `openssl`, preferably to 1.0._

I had similar problem on Arch and after a lot of googling I've managed to pinpoint the problem. As mentioned [here][http://erlang.org/pipermail/erlang-questions/2016-September/090152.html]

> Recently, new major OpenSSL version 1.1 has been released, with many
API incompatibilities, and I guess sooner or later one should port
Erlang crypto module to it. My first attempt of the porting can be
found at https://github.com/sgolovan/otp/tree/openssl11

and [here][http://erlang.2086793.n4.nabble.com/building-OTP-with-OpenSSL-1-1-0-td4718126.html]
> 2) OTP's lib/crypto/c_src/crypto.c wants "openssl/chacha.h" and "openssl/poly1305.h". which do not exist in OpenSSL 1.1.0 source. There are chacha.h and poly1305.h in crypto/include/internal, but they don't seem to be what OTP's crypto.c wants, as they don't contain required functions (like CRYPTO_chacha_20, for example)...1.0.2h, the last of OpenSSL's 1.0.x versions, simply does not have support for Chacha20/Poly1305, hence OTP compiles ok.

This issue was [solved][The crypto application now supports OpenSSL 1.1.] in Erlang itself and 20.+ will compile without any problems.  But all distributions (and I would guess any operating systems) with OpenSSL 1.1+ will cause issues while building older Erlangs (even 19.3). If you look closely into logs you will find that make is able to find OpenSSL but to to use it 
```
checking for OpenSSL >= 0.9.7 in standard locations... found; but not usable
configure: WARNING: No (usable) OpenSSL found, skipping ssl, ssh and crypto applications
```

Arch linux it

mrallen1:
Thanks for the information!

PR #246, opened 2017-11-13
circleci: add another CI provider for R15 builds

fenollp:
Follow up of discussion at https://github.com/kerl/kerl/pull/242
CircleCI is very similar to TravisCI in how simple it is to set up & how it is free for open source projects.

Build fails for some reason: https://circleci.com/gh/fenollp/kerl/17
Ideas?

PS: I'll squash when ready.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,25 @@
+version: 2
+jobs:
+  build:
+    docker:
+      - image: debian:8
+    environment:
+      - _KERL_VSN: R15B03-1
+      - KERL_CONFIGURE_DISABLE_APPLICATIONS: 'odbc'
+    steps:
+      - run: apt-get update && apt-get upgrade -y
+      - run: apt-get install -y git curl build-essential libncurses-dev libssl-dev
+      - checkout
+      - run: ./kerl update releases
+      - run:
+          command: ./kerl build "$_KERL_VSN" "$_KERL_VSN"
+          no_output_timeout: 45m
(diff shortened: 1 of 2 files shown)


fenollp:
So now circleci fails to run R15B03-1 due to crypto app not being included. https://circleci.com/gh/fenollp/kerl/19
This is exactly as described in https://github.com/kerl/kerl/issues/226

Now that debian (et al.) repos have a libssl that's too new for old Erlangs I wonder if there's even a way to build a working R15. One can downgrade libssl but package managers may upgrade it themselves later...
Maybe there's a way to install multiple libssls side by side? Maybe some libssl fork exists that is packaged in most distros happen to work with R15's build system? Last resort is building libssl from source, which might not be an issue for people that willing to run R15 anyway.

BTW I would vote to change the default behavior to fail ./configure if crypto app won't be built (instead of just showing a warning).
I would argue that very few people don't want crypto but then again probably not many want R15.
A warning is easily missed. A flag like `--dont-panic-if-no-libssl` can be added for the previous behavior.
This is kinda related to https://github.com/kerl/kerl/issues/237

mrallen1:
Where's this at now? :) Thanks.

fenollp:
This is ready. You can see the results at https://circleci.com/gh/fenollp/kerl

fenollp:
https://circleci.com/gh/kerl/kerl needs to be activated & maybe the README could show a badge for both CIs?

mrallen1:
think i activated it - a build is running anyway.  Where do you get a badge from?

mrallen1:
OK, got it.

PR #248, opened 2017-11-21
handle zlib v1.2.9+ changes on OTP 17+ - DK

akutta:
http://erlang.org/pipermail/erlang-bugs/2017-January/005215.html

Without this patch, I was not able to properly run eunit suites.  The binaries that were produced were not always being built properly.

The root issue seems to be introduced when upgrading to High Sierra.  Appears to impact other linux distributions as well, but I do not have a comprehensive list nor testing environment for this.

### DIFF
--- a/kerl
+++ b/kerl
@@ -576,6 +576,7 @@ maybe_patch_darwin()
         apply_r16_wx_ptr_patch >> "$LOGFILE"
     elif [ "$1" -ge 17 -a "$1" -le 19 ]; then
         apply_wx_ptr_patch >> "$LOGFILE"
+        apply_zlib_patch >> "$LOGFILE"
     fi
 }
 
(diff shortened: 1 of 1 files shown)


mrallen1:
Thanks for the PR - just to be clear: this issue affects Macs only and in High Sierra specifically?  I will read through the linked email chain but probably won't get to this until later this week, as it's Thanksgiving holiday in the U.S.

akutta:
Understood.  I believe it also impacts some Linux builds.  That said, linux allows downgrading of zlib whereas zlib appears to be at a deeper level of OSX.  

I have only experienced this error after upgrading to High Sierra.

mrallen1:
Reading through the email, it seems like this patch ought to be applied to Linux systems too.  So I will rework the PR a bit to accomplish that.

Thanks.

akutta:
Probably a good idea.  I did not have a system too properly test on, so
figured I'd do the least and validate it works then leave the final
decision up to you.

mrallen1:
This looks like the "official" patch commit: https://github.com/erlang/otp/commit/e27119948fc6ab28bea81019720bddaac5b655a7

akutta:
Oh even better!  Instead of handling the improperly compressed parts, ensure they compress properly.  

I'm happy to update this if you want unless you are already in the process of doing it.

mrallen1:
Thanks for the offer but I'm in the middle of it and I'm going to finish it. I'd be happy for you to comment on the PR when it's ready, though!

akutta:
No problem, just didn't want to duplicate effort.
I would be happy to comment when you are finished.

mrallen1:
Closing in favor of #258 

Thank you!

PR #251, opened 2017-12-10
Add upyun download mirror choice

zhongwencool:
Some countries(China) visit erlang.org and github very slowly, the average download speed is only 30k~40k. It makes the installation process very painful.
[UPYUN](https://github.com/upyun) make all resources under the [Erlang/download](http://www.erlang.org/download) mirror to [UPYUN/download](http://erlang.b0.upaiyun.com/download), make download become super fast.

But the mirror has  a little different from erlang.org, this make change ERLANG_DOWNLOAD_URL doesn't work.
erlang.org can visit directory information by 
`curl -f -q -L -s http://erlang.org/download`
but upyun can't visit directory information, but it provide another way for the same result: 
`curl -f -q -L -S http://erlang.b0.upaiyun.com/releases`
Download tar.gz url example: `http://erlang.b0.upaiyun.com/download/otp_src_R16B.tar.gz`

So I only make download *tar.gz from upyun, check MD5 and update releases still from erlang.org, I think this way is more safety.
#### Download from upyun
image
#### Download from erlang.org
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,6 +198,14 @@ be either a branch, a tag or a commit id that will be passed to `git checkout`:
     Building Erlang/OTP 19.2_dev from git, please wait...
     Erlang/OTP 19.2_dev from git has been successfully built
 
+Build from upyun
+----------------
+
+Some countries(China) visit erlang.org and github very slowly.
+[UPYUN](https://github.com/upyun) make all `*tar.gz` under the [Erlang/download](http://www.erlang.org/download) mirror to [UPYUN/download](http://erlang.b0.upaiyun.com/download), make download become super fast.
+
+   $ export DOWNLOAD_BACKEND="upyun"
+   $ kerl build 20.1 20.1_test
 
 Tuning
 ------
--- a/kerl
+++ b/kerl
@@ -31,6 +31,7 @@ GREP_OPTIONS=''
 
 DOCSH_GITHUB_URL="https://github.com/erszcz/docsh.git"
 ERLANG_DOWNLOAD_URL="http://www.erlang.org/download"
+UPYUN_DOWNLOAD_URL="http://erlang.b0.upaiyun.com/download"
 KERL_CONFIG_STORAGE_FILENAME=".kerl_config"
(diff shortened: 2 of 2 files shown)


mrallen1:
Hey, thanks for the PR. Is there a reason why you couldn't do something like:

```bash
$ ERLANG_DOWNLOAD_URL="http://erlang.b0.upaiyun.com/download" kerl build 20.1 20.1
```
and accomplish the same goal without requiring any code changes?

zhongwencool:
`curl -f -q -L -s http://erlang.org/download` can get release info from `erlang.org`.
but `curl -f -q -L -s http://erlang.b0.upaiyun.com/download` get 404 Not Found.
only `curl -f -q -L -s http://erlang.b0.upaiyun.com/releases` get the same result as `erlang.org`.
if I only change the Erlang_DOWNLOAD_URL, `kerl update releases` will be failed.

Maybe another way to achieve the same work. 

Add 
`ERLANG_RELEASES_URL="http://erlang.org/download"`
`ERLANG_DOWNLOAD_URL="http://erlang.org/download"`

`get_tarball_releases()` from **ERLANG_RELEASES_URL**
`tarball_download()` from **ERLANG_DOWNLOAD_URL**

then I can build without add any url into kerl.
```
export ERLANG_DOWNLOAD_URL="http://erlang.b0.upaiyun.com/download" 
export ERLANG_RELEASES_URL="http://erlang.b0.upaiyun.com/release"
kerl build 20.1 20.1
```

mrallen1:
It seems like it'd be super easy to make a wrapper script to do what you're proposing.

```bash
MYURL="https://example.com/download"

case "$1" in
    update)
         ./upstream_kerl $*
         ;;
    *)
        ERLANG_DOWNLOAD_URL="$MYURL" ./upstream_kerl $*
        ;;
esac
```

zhongwencool:
Thanks~ Your way is marvelous 👍

mrallen1:
💖

PR #269, opened 2018-04-03
Alpine support

fenollp:
`kerl` would fail when `kerl install`ing on Alpine because of Alpine's incomplete support for `ps -p`.

Note that `kerl build` outputs warnings about `dpkg` not being able to find the required/recommended packages. It seems dpkg's index is not built.
I am unsure how to fix that. They are false negatives anyway.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,7 +1,15 @@
 version: 2
+workflows:
+  version: 2
+  test-workflow:
+    jobs:
+      - debian8_R15B03-1
+      - centos6_20.2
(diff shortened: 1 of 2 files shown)


mrallen1:
When I added checking the Linux prereqs to kerl, I only expected to target distros based on Red Hat or Debian tools, and not something like Alpine or Gentoo or w/e. Sorry about the noise. Maybe we should add a flag that's like "shut up I know what I'm doing"

mrallen1:
This PR LGTM by the way.

PR #276, opened 2018-04-15
Fail if curl call fails

### LINKED ISSUE #275: If erlang.org is unavailable, curl ought to die noisily
It currently fails silently, as noted in #274.

The area which needs some attention is [this one](https://github.com/kerl/kerl/blob/master/kerl#L213-L218)

fenollp:
Fixes #275


mrallen1:
Yeah I didn't know erlang.org supported HTTPS. Good call. This looks solid. 👍

fenollp:
Seems to be the plus side of the latest downtime! Now if they could move to gh-pages or netlify.com that'd be great :)

Issue #274, opened 2018-04-14
Empty list of releases


Spoowy:
Hey Erlangers,

I'm getting this output on a fresh AWS EC2 Ubuntu 14 instance:

```
./kerl update releases

The available releases are:
```

And this as well

```
./kerl build 19.2 19.2
19.2 is not a valid Erlang/OTP release
```

What can I do to make this work? Last time it worked flawlessly.

Cheers,
Erwin

mrallen1:
Is it possible you don't have `sed` or `curl` installed?  `kerl` uses those utilities to fetch things from websites like erlang.org. (See [here for more if you're interested](https://github.com/kerl/kerl/blob/master/kerl#L213-L218))

mrallen1:
If you continue have problems, please paste in the output from `bash -x kerl update releases` on this issue.  Thanks.

Spoowy:
Thanks @mrallen1 

sed and curl are installed

This is the output I've got:

```
$ bash -x ./kerl update releases
+ unset ERL_TOP
+ KERL_VERSION=1.8.2
+ DOCSH_GITHUB_URL=https://github.com/erszcz/docsh.git
+ ERLANG_DOWNLOAD_URL=http://www.erlang.org/download
+ KERL_CONFIG_STORAGE_FILENAME=.kerl_config
+ '[' -z /home/ubuntu ']'
+ OTP_GITHUB_URL=https://github.com/erlang/otp
+ KERL_BASE_DIR=/home/ubuntu/.kerl
+ KERL_CONFIG=/home/ubuntu/.kerlrc
+ KERL_DOWNLOAD_DIR=/home/ubuntu/.kerl/archives
+ KERL_BUILD_DIR=/home/ubuntu/.kerl/builds
+ KERL_GIT_DIR=/home/ubuntu/.kerl/gits
+ '[' -n https://github.com/erlang/otp ']'
+ _OGU=https://github.com/erlang/otp
+ '[' -n '' ']'
+ '[' -n '' ']'
+ '[' -n '' ']'
+ '[' -n '' ']'
+ '[' -n '' ']'
+ '[' -n '' ']'
+ '[' -n '' ']'
+ '[' -n '' ']'
+ '[' -n '' ']'
+ '[' -n '' ']'
+ '[' -n '' ']'
+ OTP_GITHUB_URL=
+ KERL_CONFIGURE_OPTIONS=
+ KERL_CONFIGURE_APPLICATIONS=
[...]

PR #277, opened 2018-04-15
change default KERL_BUILD_BACKEND to "git" (from "tarball") after #274

fenollp:


mrallen1:
I am wary about making this change because it will break people who expect the tool to work "the way it always had."  The git build process imposes additional prereqs like autoconf, git, and automake that people might not have installed.

I think it would be better to set a date for the change over and then merge this work - like make an announcement that on/around June 1, 2018, we are going make this change, so if you have any automation that would be affected by this change, pin it to an older version of kerl or adjust the build process now.

mrallen1:
I think it's time to reconsider this and make it the default. We'll bump the version tag and make an announcement to erlang-questions ahead of time.

fenollp:
Rebased! Turns out the PR is even smaller now :)

I must have confused the hell out of these CIs cause I created & deleted the same branch on kerl/kerl instead of force-pushing on fenollp/kerl ;) 
So these should be false negatives.

fenollp:
@mrallen1 I rebased but I think this uncovered a regression with OTP 17 on Trusty WRT wxwidgets: https://app.circleci.com/jobs/github/fenollp/kerl/473/parallel-runs/0/steps/0-105 Ideas?

mrallen1:
I dunno - it says wxWidgets not found - not sure why that is.

mrallen1:
My usual approach would be to see if there's a patch in OTP somewhere upstream that we could possibly adapt or incorporate.

fenollp:
Nothing's changed except for the build backend, so maybe the things the OTP team does when releasing a tarball was not done for R17.0?

I could just drop Trusty as it is EOL by Canonical...

mrallen1:
Any OS that's EOL'd upstream of us we should drop from our test matrix IMO

fenollp:
Travis CI macOS builds seem to be failing due to https://pattern-match.com/blog/2019/01/27/erlang-how-to-safely-build-erlang-with-crypto-without-openssl-issues/ (the new openssl 1.1.x API).

fenollp:
@mrallen1 What do you think about cutting a v2 from here?

mrallen1:
Yeah seems like a good call to me. But like I said, I want to make an announcement at CodeBEAM and then we'll launch 2.0.0 with the switch to git builds as default

PR #284, opened 2018-06-16
Fixing CDPATH affecting otp_build

### LINKED ISSUE #239: MacOS with zsh build issue
I ran into an issue building erlang with MacOS (High Sierra) running zsh and oh-my-zsh. The error was:

```bash
 MAKE	depend
/bin/sh: line 0: cd: erts/emulator: No such file or directory
make: *** [depend] Error 1
```

This is caused by how zsh handles the CDPATH environment variable. If you don't have '.' in your CDPATH and you run:

```bash
cd erts/emulator
```

which the Makefile does, the cd will fail. Adding '.' to your CDPATH fixes the issue. If you don't want '.' in your CDPATH you can run:

```bash
export CDPATH=$CDPATH:.
```

and kerl builds will finish without errors.

I thought you might want to add a note in the README about this.

gullitmiranda:
fix #239 
workaround to https://github.com/erlang/otp/pull/1841

### DIFF
--- a/kerl
+++ b/kerl
@@ -24,6 +24,9 @@
 
 unset ERL_TOP
 
+# Make sure CDPATH doesn't affect cd in case path is relative.
+unset CDPATH
+
 KERL_VERSION='1.8.2'
 
 DOCSH_GITHUB_URL='https://github.com/erszcz/docsh.git'


mrallen1:
Thanks!

PR #295, opened 2018-08-03
Fix Bash completion for Bash 3

tzvetkoff:
Well, since Bash 3 lacks the `readarray` (`mapfile`), kerl's completion does not work - and, unfortunately, Яpple still ships Bash 3 as default shell on macOS.

In this change we directly assign the result of `compgen` to `COMPREPLY` instead of reading the output of `compgen` into a variable.

Tested on:
- 3.2.57(1)-release (macOS default)
- 4.4.23(1)-release (Homebrew)

### DIFF
--- a/bash_completion/kerl
+++ b/bash_completion/kerl
@@ -12,42 +12,42 @@ _kerl()
 
     case $prev in
         kerl)
-            mapfile -t COMPREPLY < <( compgen -W 'build install update list delete active path status' -- "$cur" )
+            COMPREPLY=($(compgen -W 'build install update list delete active path status' -- "$cur"))
             ;;
         list)
-            mapfile -t COMPREPLY < <( compgen -W 'releases builds installations' -- "$cur" )
+            COMPREPLY=($(compgen -W 'releases builds installations' -- "$cur"))
             ;;
         build)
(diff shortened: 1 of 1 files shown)


fenollp:
LGTM. Do you need a new tag for this?

@mrallen1 Any idea why contributor PRs are not checked by CircleCI?

tzvetkoff:
Well, yeah, ideally this needs a new patch release so it will be updated in brew / other package managers as well.

fenollp:
Could you give a try at [what shellcheck suggest here](https://circleci.com/gh/kerl/kerl/77)?

fenollp:
Meanwhile I've tagged https://github.com/kerl/kerl/releases/tag/1.8.4 so that homebrew can bring in completion again. Thanks.

tzvetkoff:
10x!

I checked what ShellCheck suggests but there's no clean way to do this with `read -a` as it reads only a single line - we'll always need to split the string.
I'd rather disable `SC2207` for `COMPREPLY=(...)` but I guess that's up to you.

fenollp:
I agree that disabling that check is the best compromise here. Could you make a PR pleaase?

tzvetkoff:
https://github.com/kerl/kerl/pull/296 :-)

PR #296, opened 2018-08-05
Disable SC2207 for Bash completion

tzvetkoff:
SC2207 suggests to use `mapfile` or `read -a`.
`mapfile` was added in Bash 4, while `read -a` only reads the first line from a file, so both are no go.


fenollp:
Thank you very much :)

PR #299, opened 2018-08-11
clone repos with the smallest depth for speed and space

fenollp:
Since kerl first does a mirror clone (to act as a cache) this optimization
cannot be used on these clones.

Needs https://github.com/kerl/kerl/pull/298 to be merged first to ensure things still work.


fenollp:
@mrallen1 Thoughts on this?

mrallen1:
Seems like a good idea as long as people are just using the checkouts to build from and not, say, as some kind of development kickstart.

fenollp:
Yes I’m annoyed by this removing the possibility of using these clones as checkouts too. 
I wanted to save some time on the first clone anyway so this didn’t really achieve what I wanted.

PR #300, opened 2018-09-08
Added a tmp_dir var

Awlexus:
I need the script to also use a custom tmp-dir, so I implemented that.

Context: I had the crazy idea to install erlang on android, where I don't have access to `/tmp`


fenollp:
Would be nice to mention the introduced env variable in the README. 
I’m curious: isn’t it easier to cross compile Erlang for Android? I have no experience crosscompiling Erlang though.

Awlexus:
Sorry, didn't think about that. I mentioned it under the section ["Using shell export command in .kerlrc"](https://github.com/Awlexus/kerl/blob/master/README.md#using-shell-export-command-in-kerlrc)

Tbh, I don't know much about cross compiling. I just wanted to give it a shot, since I have gcc installed in Termux. I honestly wish I had known about it, **before** I compiled it on my raspberry pi...

mrallen1:
Thanks for the PR and the idea.

PR #304, opened 2018-10-03
if not supplied with a build name use the release name (fixes #303)

### LINKED ISSUE #303: if not supplied with a build name use the release name
When doing

```
kerl build 20.3.2
```

kerl complains about the name of the build to be missing. 

It would be nice if kerl just picked the release name as the default build name, especially since this is what I end up doing every time anyway.

fenollp:


mrallen1:
LGTM 👍

PR #317, opened 2019-07-15
Deploy kerl to pypi and install kerl from pypi

4383:
Allow user to install kerl from [python pypi warehouse](https://pypi.org/).

These changes introduce a python packaging system for kerl.
They package kerl into a python package, deploy it to pypi and allow
users to install kerl from pypi by using:

pip install pykerl

The kerl project name is already in use on pypi so I decided to name
it pykerl.

The package version is based on the git tag version, so when you built a
a new version, the version retrieved correspond to your tags that
already existes.

The python package and deployed is built by using travis-ci. Everythings
is automatic, you just need to deploy a new git tag to your github repo
to deploy a new version on pypi.

By example if you create a tag `1.8.6` and after you push it on github
then a new package in version `1.8.6` will be available on pypi.

You don't need to bump the version manually for the python installer part.

/!\ 
**You need to create a pypi account first and to update the password section in the travis yaml file. Read  the following documentations for more informations:**
- https://docs.travis-ci.com/user/encryption-keys
- https://docs.travis-ci.com/user/deployment/pypi/

/!\

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,5 @@
 *~
+build/
+dist/
+*.egg-info
+.eggs/
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: bash
 sudo: false
 os: osx
-
 env:
   global:
   - KERL_BASE_DIR="$TMPDIR"/.kerl
@@ -14,7 +13,6 @@ env:
   - _KERL_VSN=18.3
   - _KERL_VSN=17.5
   - _KERL_VSN=R16B03-1
-
 before_script: set -e
 after_script: set +e
 script:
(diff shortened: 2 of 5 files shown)


mrallen1:
Please help me understand the rationale for making kerl available from pypi?

4383:
@mrallen1 just allow to users to install it in a different way from a different place, that's all.

You only provide install from brew. No matter if kerl come from pypi, or from elsewhere in world, the main goal of these changes is to introduce a simple alternative to brew for your users and especially all the linux users who already have pip installed by default locally.

fenollp:
I'm not against this, if `pykerl` versioning matches exactly `kerl`'s.

If we do merge this I'd like to see a `snap` of `kerl`, if that's even possible.

4383:
@fenollp the pykerl version is based and generated and deployed to pypi from your git tag so no risk there. If you push a new kerl tag to github then CI will build the corresponding version and push it to pypi.

https://github.com/kerl/kerl/pull/317/files#diff-2eeaed663bd0d25b7e608891384b7298R4

> I'm not against this, if `pykerl` versioning matches exactly `kerl`'s.
> 
> If we do merge this I'd like to see a `snap` of `kerl`, if that's even possible.

I'm far from be an erlang expert I'm a newcomer in your world, I think I understand what [`snap` is](https://snap.readthedocs.io/en/latest/), but I'm not about to understand your related sentence.

Issue #335, opened 2020-03-26
Unable to build 22.3 on macOS 10.15.4


eproxus:
On a clean install of macOS 10.15.4 with Homebrew `openssl@1.1` (1.1.1d) and `wxmac` 3.1.3 the following error is generated when building Erlang/OTP 22.3:

```
» kerl build 22.3 22.3                                                       1s
Downloading OTP-22.3.tar.gz to /Users/user/.kerl/archives
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   120  100   120    0     0    336      0 --:--:-- --:--:-- --:--:--   335
100 53.2M    0 53.2M    0     0  6954k      0 --:--:--  0:00:07 --:--:-- 9950k
Extracting source code
Building Erlang/OTP 22.3 (22.3), please wait...
APPLICATIONS DISABLED (See: /Users/user/.kerl/builds/22.3/otp_build_22.3.log)
 * jinterface     : Java compiler disabled by user
 * odbc           : ODBC library - header check failed

DOCUMENTATION INFORMATION (See: /Users/user/.kerl/builds/22.3/otp_build_22.3.log)
 * documentation  :
 *                  fop is missing.
 *                  Using fakefop to generate placeholder PDF files.

Build failed.
 RANLIB ../lib/internal/x86_64-apple-darwin19.4.0/liberts_internal_r.a
 GEN    obj/x86_64-apple-darwin19.4.0/opt/MADE
 LD     /Users/user/.kerl/builds/22.3/otp_src_22.3/bin/x86_64-apple-darwin19.4.0/beam.smp
ld: weak import of symbol '___darwin_check_fd_set_overflow' not supported because of option: -no_weak_imports for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[4]: *** [/Users/user/.kerl/builds/22.3/otp_src_22.3/bin/x86_64-apple-darwin19.4.0/beam.smp] Error 1
make[3]: *** [opt] Error 2
make[2]: *** [opt] Error 2
make[1]: *** [smp] Error 2
make: *** [emulator] Error 2

Please see /Users/user/.kerl/builds/22.3/otp_build_22.3.log for full details.
```

Full log: [otp_build_22.3.log](https://github.com/kerl/kerl/files/4387399/otp_build_22.3.log)

<details>
    <summary>Brew Packages</summary>

    ```
    antibody
[...]

PR #339, opened 2020-04-11
md improving erlang version check version check and xcode

mrcdll:
This PR improves the erlang version check and adds the check if xcode version is smaller than 11.4 before applying the `apply_catalina_no_weak_imports_patch`

### DIFF
--- a/kerl
+++ b/kerl
@@ -599,8 +599,12 @@ maybe_patch() {
     release=$(get_otp_version "$2")
     case "$1" in
         Darwin)
+            xcode_version=$(/Applications/Xcode.app/Contents/Developer/usr/bin/xcodebuild -version | head -n 1 | awk '{print $2}')
             # Catalina and clang require a "no-weaks-import" flag during build
-            if [[ $(uname -r) == "19"* ]] && [[ "$release" -lt 23 ]] && [[ $2 != "22.3.1" ]]; then
+            if [[ $(uname -r) == "19"* ]] && \
+               [[ "$release" -lt 23 ]] && \
+               [[ $(printf "$2\n22.3.1" | sort | head -n 1) != "22.3.1" ]] && \
+               [[ $(printf "$xcode_version\n11.4" | sort | head -n 1) == "11.4" ]]; then
                 apply_catalina_no_weak_imports_patch >>"$LOGFILE"
             fi
             maybe_patch_darwin "$release"
--- a/kerl
+++ b/kerl
@@ -599,17 +599,9 @@ maybe_patch() {
     release=$(get_otp_version "$2")
     case "$1" in
         Darwin)
-            xcode_version=$(/Applications/Xcode.app/Contents/Developer/usr/bin/xcodebuild -version | head -n 1 | awk '{print $2}')
-            command_line_tools_version=$(xcode-select -v | awk '{print $3}' | sed 's/.$//')
-            # Catalina and clang require a "no-weaks-import" flag during build
-            if [[ $(uname -r) == "19"* ]] && \
-               [[ "$release" -lt 23 ]] && \
-               [[ $(printf "$2\n22.3.1" | sort --version-sort | head -n 1) != "22.3.1" ]] && \
-               [[ $(printf "$xcode_version\n11.4" | sort --version-sort | head -n 1) == "11.4" ]] && \
-               [[ "$command_line_tools_version" -le 2373 ]]; then
-                apply_catalina_no_weak_imports_patch >>"$LOGFILE"
-            fi
             maybe_patch_darwin "$release"
+            # Catalina and clang require a "no-weaks-import" flag during build
+            maybe_patch_catalina "$release" "$2"
             ;;
         SunOS)
             maybe_patch_sunos "$release"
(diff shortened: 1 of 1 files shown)


mrcdll:
@mrallen1 this should be able to cover the conditions described here https://github.com/kerl/kerl/pull/337#issuecomment-611597487 and here https://github.com/kerl/kerl/issues/335#issuecomment-611431833

eproxus:
@mrcdll Note that, according to my testing, you need to check for command line tools as well. I downgraded only Xcode to 11.3 and that produced the same build error, until I also downgraded Command Line Tools for Xcode to 11.3.1.

kerl
```diff
@@ -599,8 +599,12 @@ maybe_patch() {
     release=$(get_otp_version "$2")
     case "$1" in
         Darwin)
+            xcode_version=$(/Applications/Xcode.app/Contents/Developer/usr/bin/xcodebuild -version | head -n 1 | awk '{print $2}')
             # Catalina and clang require a "no-weaks-import" flag during build
-            if [[ $(uname -r) == "19"* ]] && [[ "$release" -lt 23 ]] && [[ $2 != "22.3.1" ]]; then
+            if [[ $(uname -r) == "19"* ]] && \
+               [[ "$release" -lt 23 ]] && \
+               [[ $(printf "$2\n22.3.1" | sort | head -n 1) != "22.3.1" ]] && \
+               [[ $(printf "$xcode_version\n11.4" | sort | head -n 1) == "11.4" ]]; then
```
fenollp, line 607: This `sort` is not semver sort. Not sure if that is an issue here but if a less ambiguous tool can be used here that'd be great.

mrcdll: @fenollp thank you for the feedback, looks like the system sort has a --version-sort that would solve it, it also handles rc versions
```
-V, --version-sort
             Sort version numbers.  The input lines are treated as file names in form PREFIX VERSION SUFFIX, where SUFFIX matches the
             regular expression "(.([A-Za-z~][A-Za-z0-9~]*)?)*".  The files are compared by their prefixes and versions (leading zeros
             are ignored in version numbers, see example below).  If an input string does not match the pattern, then it is compared
             using the byte compare function.  All string comparisons are performed in C locale, the locale environment setting is
             ignored.
```

mrcdll:
> @mrcdll Note that, according to my testing, you need to check for command line tools as well. I downgraded only Xcode to 11.3 and that produced the same build error, until I also downgraded Command Line Tools for Xcode to 11.3.1.

Hey @eproxus thank you for the detail. Can I ask you where did you find the xcode command line tools version? This is my result.
```
❯ xcode-select -v
xcode-select version 2373.
```

kerl
fenollp: Oh I was under the impression `sort -V` wasn't available on osx. Nice.

eproxus:
@mrcdll I didn't specifically check for the version, I just knew I hade Command Line Tools for Xcode 11.4 installed, and manually uninstalled and downgraded to 11.3.1. I'm also on version `2373` when running `xcode-select`.

mrcdll:
@eproxus  `xcode-select` is the command-line tools from Mac. Apparently it doesn't have a semantic version, but we need to test against that number, do you know which version is working for you?

eproxus:
As far as I can tell it's version `2373` (just reinstalled version CLT 11.3.1 and got this).

mrcdll:
@eproxus and @fenollp this last commit should address both your comments.

eproxus:
Can't say for sure, but I also have a hunch the version sort flag is pretty new in macOS. I remember dealing with complex `sort -n -k...` flags before to do version sort. Can't find any reliable info on this though. Anyone have pre-Catalina machines to test on?

mrcdll:
technically `sort --version-sort` it should never run unless is on Catalina
```
[[ $(uname -r) == "19"* ]]
```
will short circuit and prevent the rest to run

mrallen1:
I would prefer to see these version checks put into their own functions... would you mind revising your PR to create functions to check xcode version, and more fine-grained OTP release parsing functions so we can examine the minor release numbers - example `otp_release 22.3.2` gives 22, but now we ALSO need to check the minor release `3.2` too... so we should have function to return those components.

mrcdll:
Hey @mrallen1 thank you for the feedback. I believe this check provides verification for the minor release and it works with -rc as well. Did you have in mind a different implementation for it?
```
[[ $(printf "$2\n22.3.1" | sort --version-sort | head -n 1) != "22.3.1" ]] && \
```

mrallen1:
No that seems fine - not sure how portable the `--version-sort` flag is but... it's ok for now. If we to figure out something more portable we can cross that bridge later.

mrcdll:
`--version-sort` is present in every Catalina version, which is the only case this block would run

mrallen1:
Yeah that's fine for now I guess. What I want is a function

```bash
otp_release_minor() {
    $rel = $1
    # get the minor release id from otp release
    # example: 20.3.3 returns 3, 21.2 returns 2, 22.0 returns 0
    # implementation TBD

    return $answer
}
otp_release_release() {
    $rel = $1
    # get the release id from the otp release
    # example 20.3.2 returns 2, 22.2.1 returns 1, 22.0 returns 0
    # implementation TBD
    return $answer
}

if [ otp_release_release "22.3.1" -ge 1 ];
   # should be true
   echo "true"
else
   echo "false"
fi
```

That's how I'd _like_ to see these checks written.

mrcdll:
hey @mrallen1 ! I refactored the condition checks for patching Catalina, I believe what we have right now is robust for the use case. I was planning to implement your semver checker in the following iteration. Sounds good to you?

mrallen1:
Yeah, I like this better. Thanks!!

PR #344, opened 2020-07-03
Allow multiple KERL_DOC_TARGETS

### LINKED ISSUE #341: Can't choose more then one doc target
In https://github.com/kerl/kerl/pull/336 support for building the new doc chunks were added by setting `DOC_TARGETS` when building the docs. This is great, and even better you can tweak said `DOC_TARGETS` using `KERL_DOC_TARGETS`.

I like having the man pages around, and also be able to use `h/1` in the shell would be nice. Unfortunately you can only build one such doc target, as the string concatenation here:

https://github.com/kerl/kerl/blob/cd6280cb7de71e0d3c3c61648aeca12e2b70cffb/kerl#L947-L948

Makes the build fail with ``make[3]: *** No rule to make target `chunks', needed by `docs'.  Stop.``, when `KERL_DOC_TARGETS="man chunks"`.

Manually running `make release_docs DOC_TARGETS="man chunks"` works like a charm though.

The solution is probably to avoid string concatenation altogether. I'm thinking either writing out the command in full every time it's run (maybe another small helper function?), or setting `DOC_TARGETS` as an exported environmental variable and hope `make` picks it up. The latter _should_ work on most systems, but if kerl wants to support stuff like AIX/Sun/Solaris/... then it might be a solution.

What do you think?

maxno-kivra:
Fixes #341

### DIFF
--- a/kerl
+++ b/kerl
@@ -943,16 +943,15 @@ _do_build() {
     fi
     if [ -n "$KERL_BUILD_DOCS" ]; then
         echo 'Building docs...'
-        release=$(get_otp_version "$2")
-        cmd="make docs DOC_TARGETS=$KERL_DOC_TARGETS"
-        release_cmd="make release_docs DOC_TARGETS=$KERL_DOC_TARGETS RELEASE_ROOT=$KERL_BUILD_DIR/$2/release_$1"
+        export DOC_TARGETS="$KERL_DOC_TARGETS"
+        export RELEASE_ROOT="$KERL_BUILD_DIR/$2/release_$1"
 
-        if ! $cmd >>"$LOGFILE" 2>&1; then
+        if ! make docs >>"$LOGFILE" 2>&1; then
             show_logfile 'Building docs failed.' "$LOGFILE"
             list_remove builds "$1 $2"
             exit 1
         fi
-        if ! $release_cmd >>"$LOGFILE" 2>&1; then
+        if ! make release_docs >>"$LOGFILE" 2>&1; then
             show_logfile 'Release of docs failed.' "$LOGFILE"
             list_remove builds "$1 $2"
             exit 1
@@ -1262,8 +1261,9 @@ rehash
 ACTIVATE_CSH
 
     if [ -n "$KERL_BUILD_DOCS" ]; then
-        release_docs="make release_docs DOC_TARGETS=$KERL_DOC_TARGETS RELEASE_ROOT=$absdir"
-        if ! (ERL_TOP="$ERL_TOP" $release_docs >/dev/null 2>&1); then
+        export DOC_TARGETS="$KERL_DOC_TARGETS"
+        export RELEASE_ROOT="$absdir"
+        if ! (ERL_TOP="$ERL_TOP" make release_docs >/dev/null 2>&1); then
             echo "Couldn't install docs for Erlang/OTP $rel ($1) in $absdir"
             exit 1
         fi


maxno-kivra:
@mrallen1 what do you think?

kerl
```diff
@@ -943,16 +943,15 @@ _do_build() {
     fi
     if [ -n "$KERL_BUILD_DOCS" ]; then
         echo 'Building docs...'
-        release=$(get_otp_version "$2")
-        cmd="make docs DOC_TARGETS=$KERL_DOC_TARGETS"
-        release_cmd="make release_docs DOC_TARGETS=$KERL_DOC_TARGETS RELEASE_ROOT=$KERL_BUILD_DIR/$2/release_$1"
```
mrallen1, line 948: Isn't the main issue that this line needs to have quotes?

maxno-kivra: I tried to inline this command in the `if` statement, but it didn't work. For some reason the `make DOC_TARGETS="$KERL_DOC_TARGETS"` still split it and failed with `missing target chunks`.

Exporting does work nicely though.

maxno-kivra:
Ping @mrallen1. This should be good to go, yes?

mrallen1:
Hi - thanks for the patch and I'm sorry it's taken me so long to do anything about this. I am working on a branch to fix this in a different way than using an exported variable.

maxno-kivra:
Why is exporting `DOC_TARGETS`/`RELEASE_ROOT` an issue?

mrallen1:
> Why is exporting `DOC_TARGETS`/`RELEASE_ROOT` an issue?

Because kerl gets used in all kinds of automation contexts especially for building Erlang and people might have exported their own values for those variables on their build machines/environments/CI/docker containers. 

Kerl's own shell variables are prefixed with `KERL_` so I feel like we can do whatever we want with them, but `DOC_TARGETS` and so on are OTP variables and we shouldn't export them from inside kerl especially when we can achieve the same result without doing so.

maxno-kivra:
> Because kerl gets used in all kinds of automation contexts especially for building Erlang and people might have exported their own values for those variables on their build machines/environments/CI/docker containers.

I wholeheartedly agree that we must stay backwards compatible, and ensure peoples setup doesn't break. But, `kerl` isn't sourced, right? Which means any exported variables stays inside, and won't mess with peoples setup.

> Kerl's own shell variables are prefixed with `KERL_` so I feel like we can do whatever we want with them, but `DOC_TARGETS` and so on are OTP variables and we shouldn't export them from inside kerl especially when we can achieve the same result without doing so.

The OTP ones are overwritten anyways, so it doesn't matter if the user has exported them or not. If this should be changed to make kerl inherit these, then I assume this needs to be updated:

https://github.com/kerl/kerl/blob/490c077fd110818d2052a5ecafa47e5db44cc09c/kerl#L85-L86

Perhaps the cleanest way would be to export all the OTP environmental variables upfront, while unifying the `KERL_` prefixed ones with the ones from OTP? This should make `kerl` more robust if OTP changes a `make` command to start depending on some of the other ones. For example `RELEASE_ROOT`.

mrallen1:
They've overwritten _while someone is using kerl_ - but again - this tool gets used in all sorts of automation contexts which we don't know about. So I think being conservative about exports (especially exports defined by the upstream project) is the right behavior.

maxnordlund:
> They've overwritten _while someone is using kerl_

You mean inside the kerl process while it is running? Unless the user sources kerl, it can't affect the users environment, and kerl isn't supposed to be sourced (except the activate script), right? Or am I missing something?

> - but again - this tool gets used in all sorts of automation contexts which we don't know about. So I think being conservative about exports (especially exports defined by the upstream project) is the right behavior.

I totally agree, the more robust the better. I also think a user of kerl would be surprised if setting _both_ OTP and `KERL_` environmental variables affected the build, but only for some steps of the build.

kerl must overwrites/passe along all the variables used by OTP at each step, e.g. make call, or OTP environmental variables will bypass kerl. I'm not sure if you can do something like that right now, but before `KERL_DOC_TARGETS` was added, you could tweak OTP by setting `DOC_TARGETS`. Also, if OTP changes any of it's steps to use another variable, kerl would subtle break. [Although I'm sure someone is/will be using this](https://xkcd.com/1172/)

Given that, I think setting the OTP variables upfront would prevent surprises like that. This would reduce the maintenance burden to export a new OTP variable if they add one. Instead of updating every call to make.

___

Basically I fail to see how exporting or passing variables can affect the user. It's an implementation detail, right?

mrallen1:
> Basically I fail to see how exporting or passing variables can affect the user. It's an implementation detail, right?

No, it can affect build environments in unexpected and troublesome ways. Exporting environment variables from upstream into a shell environment should be avoided whenever possible.

maxnordlund:
I fail to see how modifying environmental variables _inside_ a process can affect the parent process. I also think there's more nuance to using upstream, in our case Erlang/OTP, environmental variables. For example [homebrew](https://brew.sh/) removes some variables precisely because they can cause weird errors.

https://github.com/Homebrew/homebrew-core/blob/8e2108774bcb5a7057b55a97b8fcd22300d065c1/Formula/erlang.rb#L37-L39

I think we're not going to convince each other otherwise, so maybe just agree to disagree. I'm also not going to stop a bugfix, so if your alternative solution works, great 😄 Once verified, you can close this.

If in the future we get a weird error from OTP, then maybe this discussion can help us remember to look at Erlang/OTP environmental variables.

PR #345, opened 2020-07-05
Use POSIX -path flag instead of GNU -wholename flag for find.

juster:
Here is a minor patch to the activate script. The activate script was producing an error message on OpenBSD 6.6. The error was:

> find: -wholename: unknown option

`-path` is equivalent to `-wholename` according to the GNU doc here:

https://www.gnu.org/software/findutils/manual/html_mono/find.html#Full-Name-Patterns

Thanks,
Justin


mrallen1:
Cool thanks

PR #355, opened 2020-10-19
Update to build on macosx catalina

raghavkarol:
Remove reference to xcodebuild and dependency on Xcode. It's not a good idea to depend on Xcode for building OTP with kerl since we only needed clang and friends that are part of Xcode command line tools.

Tested with this configuration.

- OSX 10.15.7
- OTP 21.3.8.4
- Xcode Command line Tools 12

### DIFF
--- a/kerl
+++ b/kerl
@@ -27,7 +27,7 @@ unset ERL_TOP
 # Make sure CDPATH doesn't affect cd in case path is relative.
 unset CDPATH
 
-KERL_VERSION='2.0.0'
+KERL_VERSION='2.0.1'
 
 DOCSH_GITHUB_URL='https://github.com/erszcz/docsh.git'
 ERLANG_DOWNLOAD_URL='https://www.erlang.org/download'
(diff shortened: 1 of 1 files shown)


motobob (approved):
YOHOO!

motobob:
@mrallen1 would be great to get this merged and out.

kerl
```diff
@@ -642,6 +642,10 @@ maybe_patch_all() {
     if [ "$1" -ge 17 ] && [ "$1" -le 19 ]; then
         apply_zlib_patch >> "$LOGFILE"
     fi
+
+    if [ "$1" -ge 22 ] && [ "$1" -le 23 ]; then
```
mrallen1, line 646: As far as I can tell, this patch **only** applies to mac OS - can you put this in `maybe_patch_darwin()` function which is around line 651?

kerl
```diff
@@ -27,7 +27,7 @@ unset ERL_TOP
 # Make sure CDPATH doesn't affect cd in case path is relative.
 unset CDPATH
 
-KERL_VERSION='2.0.0'
+KERL_VERSION='2.0.1'
```
mrallen1, line 30: Please don't modify this - I do that myself when I roll a new release. Thanks.

mrallen1 (changes requested):
If we can get the changes done, I'll merge and release it. Thanks.

mrallen1 (approved):

mrallen1:
I'm going to wait for the tests to run and then I'll get this merged. Thanks!!

raghavkarol:
Thank you also, see you around :)

mrallen1:
FYI @motobob @raghavkarol the `2.0.1` tag has been pushed to the repo or you can just pull for the update. Thanks again.

Issue #346, opened 2020-08-10
Configure error with erlang 23.0.3 in MacOS 11 (Big sur)


FGMGIT:
Hello if unfortunately or necessarily you are working with MacOS 11 Big Sur. And you try uninstall brew erlang to reinstall it seems that you will not be able to reinstall that version with asdf, kerl or brew, even building Erlang / OTP from scratch will not be possible for OTP 23.0.3.

It seems that the only version that builds for MacOSX 11 is Erlang/OTP24. With 23.0.3 you will have:

````
checking OTP version... 23.0.3
configure: error: 

  You are natively building Erlang/OTP for a later version of MacOSX
  than current version (11.0). You either need to
  cross-build Erlang/OTP, or set the environment variable
  MACOSX_DEPLOYMENT_TARGET to 11.0 (or a lower version).
````
Try to set MACOSX_DEPLOYMENT_TARGET to 11.0 or 10.5 doesn't seems to work 

`export MACOSX_DEPLOYMENT_TARGET=11`

I have only been able to build Erlang on MacOSX 11 by following this bug note. But only in OTP 24.

https://bugs.erlang.org/browse/ERL-1305 

and erlang direct installation:

https://github.com/erlang/otp

Is there any advance with support for MacOS 11? or someone facing problems with erlang 23.0.3 in MacOS 11?

UPDATE!!! To compile [Erlang 23.0.3](https://github.com/erlang/otp/tree/maint) from source yo can use "maint" brach, they rebased bugs related with MacOS 11 builds into that branch.

**Comment by luos** [comment]
Hi,

The problem comes from the `configure.in` file found in most Erlang releases, it is pretty easy fix for me it resolved building 21.3 on Big Sur: 

1. Try a build, it will fail:
```
$ kerl build 21.3 21.3
Extracting source code
Building Erlang/OTP 21.3 (21.3), please wait...
Configure failed.
checking OTP release... 21
checking OTP version... 21.3
configure: error:

  You are natively building Erlang/OTP for a later version of MacOSX
  than current version (11.0.1). You either need to
  cross-build Erlang/OTP, or set the environment variable
  MACOSX_DEPLOYMENT_TARGET to 11.0.1 (or a lower version).

Please see /Users/x/.kerl/builds/21.3/otp_build_21.3.log for full details.

```

2. Go and edit the file `~/.kerl/builds/21.3/otp_src_21.3/configure.in` and look for the following line, it should be around line 400:

```
[...]

PR #356, opened 2020-12-10
Apply Erlang patch to build on OSX Big Sur

fhunleth:
This applies https://github.com/erlang/otp/commit/f1044ef9e35da26f276b8127640e177d67aade6a which currently exists in `master` to fix an error when configuring the build on OSX Big Sur.

See https://github.com/erlang/otp/pull/2871 for details.

Fixes https://github.com/kerl/kerl/issues/346.

Tested with this configuration.

  - OSX 11.0.1
  - OTP 23.1.4 and 23.1.5


mrallen1 (approved):
lgtm - thanks so much

PR #358, opened 2021-01-04
Replace Big Sur patch with sed-based bypass for installing OTP 21, 22, and 23

OldhamMade:
The patch added in #356 (to release 2.0.2) only applies to OTP 23, and doesn't work with earlier OTP versions. This PR provides an alternative approach.

This version has been tested with the latest versions of OTP 23, 22, and 21. I've also tested with OTP 20, but the build fails at a later point for (from what I can see) unrelated reasons.

### DIFF
--- a/kerl
+++ b/kerl
@@ -737,7 +737,7 @@ maybe_patch_bigsur() {
     release="$1"
     if is_osx_bigsur && \
         [[ "$release" -lt 24 ]]; then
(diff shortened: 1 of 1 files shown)


joedevivo:
👍 Worked for me on BigSur & 22.0.3

taras-vi:
Doesn't work with 22.3

PR #363, opened 2021-02-02
Add `\` to commands to avoid system aliases

JoaoFukuda:


mrallen1:
Thanks. Going to let CI run through this change, but planning to merge it.

JoaoFukuda:
I just had to interactively rebase and force push because there was some other commands I wanted escaped and didn't wanted to make a new commit for, so I reused the last commit.

PR #373, opened 2021-06-09
Add KERL_DOC_TYPES build configuration explanation.

iporsut:
Add the explanation of the default behaviour of `KERL_DOC_TYPES` and how to configure it.


mrallen1:
Thanks!

PR #377, opened 2021-09-22
Add pacman check to avoid false warnings

### LINKED ISSUE #375: Lot of false warnings on Arch Linux with dpkg package installed
I am getting these warnings:

```
WARNING: It appears that a required development package 'libssl-dev' is not installed.
WARNING: It appears that a required development package 'make' is not installed.
WARNING: It appears that a required development package 'automake' is not installed.
WARNING: It appears that a required development package 'autoconf' is not installed.
WARNING: It appears that a required development package 'libncurses5-dev' is not installed.
WARNING: It appears that a required development package 'gcc' is not installed.
```

The reason is that I am on Arch Linux and installed [dpkg](https://archlinux.org/packages/community/x86_64/dpkg/) package which allows me to install `.deb` archives but most of the packages are installed through native package manager `pacman`.

Could the mechanism for checking used package manager be more strict? I get that I can ignore the warnings but they are annoying and false.

Sgiath:
Fixes #375

### DIFF
--- a/kerl
+++ b/kerl
@@ -428,11 +428,12 @@ assert_build_name_unused() {
 _check_required_pkgs() {
     has_dpkg=$(command -v dpkg)
     has_rpm=$(command -v rpm)
+    has_pacman=$(command -v pacman)
     if [ -n "$has_dpkg" ] || [ -n "$has_rpm" ]; then
         # found either dpkg or rpm (or maybe even both!)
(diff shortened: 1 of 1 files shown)


jadeallenx:
There has to be a better way to test for various package managers (see #439 for a proposal)

paulo-ferraz-oliveira:
If #471 lands, this should be easier to accomplish, and also more flexible.

PR #380, opened 2021-09-24
Pin Homebrew OpenSSL version to @1.1

moogle19:
Since the release of OpenSSL 3.0 Homebrew defaults to that version, but Erlang/OTP currenly doesn't support OpenSSL 3.0.

This either results in a build failure (when `@3` is installed) or in Erlang/OTP to be build without ssl/crypto support (when `@3` is not installed). 

Pinning the version to `@1.1` fixes the problem, until Erlang/OTP support OpenSSL 3.0

### DIFF
--- a/kerl
+++ b/kerl
@@ -827,7 +827,7 @@ _do_build() {
             if ! echo "$KERL_CONFIGURE_OPTIONS" | \grep 'with-ssl' >/dev/null 2>&1; then
                 whichbrew=$(command -v brew)
                 if [ -n "$whichbrew" ] && [ -x "$whichbrew" ]; then
-                    brew_prefix=$(brew --prefix openssl)
+                    brew_prefix=$(brew --prefix openssl@1.1)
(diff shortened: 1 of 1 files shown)


paulo-ferraz-oliveira:
Have you considered `--with-ssl=/usr/local/opt/openssl@1.1` in `KERL_CONFIGURE_OPTIONS`? I know what you're changing is the default, but this might be helpful in the meantime. I can at least build 22 to 24 with it. 👍

moogle19:
@paulo-ferraz-oliveira 
Thanks for the tip. I ended up linking `openssl@3` to `openssl@1.1` to build `24.1`,  but you approach seems cleaner.

alan-andrade:
I'm running into this after a couple hours of not being able to install erlang 24.1. Thanks for posting a solution.

alan-andrade (approved):
I'm not sure why I can approve but I think this would've saved me 2 hours today. It should go in for other people.

jadeallenx:
OK, sold. Thank you!

PR #385, opened 2021-11-01
Always run autoconf when configure files have been patches

### LINKED ISSUE #384: kerl 2.2.1 fails to build OTP < 23.2 on macOS Big Sur and above due to regression with autoconf patches
We recently discovered that kerl 2.2.1 fails to build older versions of OTP < 23.2 on macOS Big Sur, here's an example using asdf with the latest version of asdf-erlang and OTP 22.3.1:

```sh-session
$ asdf install erlang 22.3.1
No build named asdf_22.3.1
Downloading 22.3.1 to /Users/user/.asdf/plugins/erlang/kerl-home/archives...
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   623  100   623    0     0   3893      0 --:--:-- --:--:-- --:--:--  3893
100 87.1M  100 87.1M    0     0  9563k      0  0:00:09  0:00:09 --:--:-- 9743k
Extracting source code
Building Erlang/OTP 22.3.1 (asdf_22.3.1), please wait...
Build failed.

  You are natively building Erlang/OTP for a later version of MacOSX
  than current version (11.4). You either need to
  cross-build Erlang/OTP, or set the environment variable
  MACOSX_DEPLOYMENT_TARGET to 11.4 (or a lower version).


ERROR: /Users/user/.asdf/plugins/erlang/kerl-home/builds/asdf_22.3.1/otp_src_22.3.1/make/configure failed!
make: *** No rule to make target `is_cross_configured'.  Stop.
make: *** No rule to make target `all'.  Stop.

Please see /Users/user/.asdf/plugins/erlang/kerl-home/builds/asdf_22.3.1/otp_build_22.3.1.log for full details.
```

This looks like a regression of #346, which was fixed earlier in 622e66eb789268f40d2d43e5ae23ab86c01f9bbd. I was able to see that the patch was still applied correctly to the autoconf files. However, I tracked this regression down to the recent change in b5b38ccc559d45deb2f47c35b56fad4591cadecf, which now skips autoconf when building from prebuilt source tarballs, and the patched autoconf files end up having no effect on the included configure scripts.

I can think of two approaches to fix this:
1. patch the generated configure scripts too, or
2. re-enable autoconf when building from tarballs.

In case anyone else using asdf stumbles upon this, there's a workaround that downgrades to an older version of asdf-erlang, which in turn uses an older version of kerl:

```sh-session
$ asdf plugin-update erlang 26321a566fc6f59d3edbdce6e737d7bf469df845
$ brew install autoconf@2.69 openssl@1.1
$ brew link --overwrite autoconf@2.69
$ KERL_CONFIGURE_OPTIONS="--with-ssl=$(brew --prefix openssl@1.1)" asdf install
```

Using a specific version of autoconf works around #369, and the OpenSSL version is required due to #380 since that fix isn't included in kerl 2.2.0.

garazdawi:
Closes #384

### DIFF
--- a/kerl
+++ b/kerl
@@ -656,6 +656,7 @@ maybe_patch_darwin() {
         apply_wx_ptr_patch >>"$LOGFILE"
     elif [ "$1" -ge 21 ] && [ "$1" -le 23 ]; then
         apply_in6addr_test_patch >> "$LOGFILE"
+        KERL_USE_AUTOCONF=1
     fi
 }
 
@@ -671,6 +672,7 @@ maybe_patch_catalina() {
         compare_sem_version $clang_version ">" "11.4" && \
         [[ "$command_line_tools_version" -le 2373 ]]; then
         apply_catalina_no_weak_imports_patch >>"$LOGFILE"
+        KERL_USE_AUTOCONF=1
     fi
 }
 
@@ -738,6 +740,7 @@ maybe_patch_bigsur() {
     if is_osx_bigsur && \
         [[ "$release" -lt 24 ]]; then
         apply_bigsur_bypass_version_check >>"$LOGFILE"
+        KERL_USE_AUTOCONF=1
     fi
 }
 


garazdawi:
I wish there was a way to write a test for this, but you need to pay for os x machines on circle ci. Might be solvable if the CI would move to github actions, though that is a bunch of work to do just to get tests for os x.

jadeallenx:
Thanks

PR #388, opened 2021-11-23
Use '-path' instead of '-wholename' in find

### LINKED ISSUE #374: Kerl - Activate script on OpenBSD 6.9
Hello!
Kerl is great, it helped me build erlang/otp for OpenBSD 6.9 so easily.
It told me I needed to install GNU-make which I easily did with pkg_add ...

Now, when trying to activate the build with ".  ~/kerl/23.3.4.5/activate", I got this error: find: -wholename: unknown option

I read the man page for GNU find and saw that "-path" could be a more portable option to "-wholename", so I went ahead and changed the activate script to use -path. So far it works, but please let me know whether it could create a problem for me or other OpenBSD users.

Oscar.

renatoaguiar:
Replace the GNU-specific '-wholename' flag in find with the portable
'-path'.

Closes #374

### DIFF
--- a/kerl
+++ b/kerl
@@ -1083,7 +1083,7 @@ REBAR_PLT_DIR="$absdir"
 export REBAR_PLT_DIR
 _KERL_ACTIVE_DIR="$absdir"
 export _KERL_ACTIVE_DIR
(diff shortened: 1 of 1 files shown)


jadeallenx:
Thanks

PR #389, opened 2021-11-25
Do not let default values override parameters set in config file

chilikk:
Fixes a bug where setting `KERL_DOC_TARGETS=man` in `.kerlrc` would be ignored due to the default value taking preference.


jadeallenx:
Thanks

Issue #390, opened 2021-12-06
Activate.fish crashes.


mmzeeman:
activate.fish crashes.

```
> .  ~/kerl/24.1.2/activate.fish
fish: Unknown command: \clind
~/kerl/24.1.2/activate.fish (line 1): 
\find "/Users/mmzeeman/kerl/24.1.2" -type d -path "*erl_interface*/bin"
^
in command substitution
	called on line 64 of file ~/kerl/24.1.2/activate.fish
from sourcing file ~/kerl/24.1.2/activate.fish
~/kerl/24.1.2/activate.fish (line 64): Unknown command
set -x _KERL_ERL_CALL_REMOVABLE (\find "/Users/mmzeeman/kerl/24.1.2" -type d -path "*erl_interface*/bin")
                                ^
from sourcing file ~/kerl/24.1.2/activate.fish
```

##### versions:

> kerl version 
2.2.3

> fish --version
fish, version 3.3.1

joshuayoerger:
I believe this is fixed by removing the backslash preceding the `find` command on line 64. At least it's working for me so far.

Line 64 should read:
```fish
set -x _KERL_ERL_CALL_REMOVABLE (find "/Users/mmzeeman/kerl/24.1.2" -type d -path "*erl_interface*/bin")
```

PR #392, opened 2021-12-17
Correct breaking activate.fish syntax error

joshuayoerger:
Resolves https://github.com/kerl/kerl/issues/390


jadeallenx:
Thanks

PR #397, opened 2022-02-03
Remove $ that prevent executing copied commands (also some whitespace)

meangrape:
The Downloading section has 3 $ for brew, curl, and chmod. If you copy
those lines with the code-block copy button you paste a dollar sign into
your command-line preventing you from running it.

Also my editor caught 3 instances of trailing whitespace.


jadeallenx (approved):
thanks

Issue #338, opened 2020-04-06
Kerl should capture install command failures


jadeallenx:
Suggested [here][1] by @thoughtarray:

Instead of discarding installation error messages, kerl should instead capture them and check for installation command success. If the installation command is unsuccessful, kerl should emit the appropriate error(s).

[1]: https://github.com/kerl/kerl/issues/289#issuecomment-610001094

paulo-ferraz-oliveira:
These days we have an install log. I'm not sure we're eating up any relevant messages any more.

The relevant part from the linked comment's issue (initial description) points to https://github.com/kerl/kerl/blob/799d5458ac3f64c8517bbcaacfe3c29d6c82beb2/kerl#L816.

These days we do that as

```
    if ! log_install_cmd "ERL_TOP=$ERL_TOP ./otp_build release -a $absdir && cd $absdir && ./Install $INSTALL_OPT $absdir"; then
        show_install_logfile "install of Erlang/OTP $rel ($build_name), in $absdir, failed!"
        exit_install
    fi
```

without filtering, so the install failures should all be visible.

I'm tentatively closing this. Feel free to reopen if there's something I've missed.

PR #401, opened 2022-03-20
Separate data and messages in output

crownedgrouse:
This PR separate data (still on stdout) and messages or errors on stderr, instead all in stdout.

If output is not a terminal (a pipe), stderr is mute. This permit better integration in automation scripts.

Changing to a minor version as it could cause incompatibility in already existing scripts in community using  kerl.


crownedgrouse:
See also #338

jadeallenx (approved):
Thanks

jadeallenx:
Rebased as #403

PR #409, opened 2022-04-24
Colorize kerl

### LINKED ISSUE #186: Consider quoting SHML in kerl so output looks a bit nicer
MIT licensed, SHML collects some VT100 escape sequences that provide terminal attributes like colors, bold, underlining, text block inversion, etc. 

https://github.com/odb/shml/blob/master/shml.sh

crownedgrouse:
Colorize stderr with colors, if ncurses available, otherwise not.
Variables KERL_COLOR_* has been added to configure this.

Fix #186


jadeallenx (approved):
Yeah this is cool, thanks!

kerl
```diff
@@ -33,18 +33,59 @@ DOCSH_GITHUB_URL='https://github.com/erszcz/docsh.git'
...
 # Redirect to stderr only if it is a terminal, otherwise mute
 stderr()
 {
 if [ -t 1 ] ; then
-    echo "$*" 1>&2
+    if [ -z "$l" ] || [ $KERL_COLORIZE -eq 0 ]; then
+        echo "$*" 1>&2
+    else
+        left="$(colorize $l)"
+        right="$(tput setaf ${KERL_COLOR_D:=9})"
```
garazdawi, line 54: @crownedgrouse Why is KERL_COLOR_D used here? Shouldn't it rather be `tput sgr0`?

crownedgrouse: 9 is the default color of the terminal, but sgr0 could be an option. 
KERL_COLOR_D can be set to this value in your env.  Feel free to do a PR to change this.

crownedgrouse: intention was to let extra display (not stderr) to be colored with a specific color between stderrs. using sgr0 do not allow this, but this not fundamental.

crownedgrouse: replacing default 9 by tput sgr0 could do anyway the trick and allowing to color inbetween by setting KERL_COLOR_D anyway.

garazdawi: Are you sure that 9 is the default? If I do `tput setaf 9` in my terminal it returns `\e[91m` which is a bright red color.

garazdawi: Seems like what 9 means varies with what `TERM` is set to. If I do:

```erlang
Eshell V13.0  (abort with ^G)
1> os:cmd("TERM=xterm tput setaf 9").
"\e[39m"
2> os:cmd("TERM=xterm-256color tput setaf 9").
"\e[91m"
```

paulo-ferraz-oliveira:
Sorry if this is not the correct issue, but although I see colors in the output (and they're Ok to me), I also see a `-n` (at the beginning of the lines), that I don't think was there before.

e.g.

```
-n local kerl found (/usr/local/bin/kerl) at version 2.5.0.
-n remote kerl found at version 2.5.0.
No upgrade required.
-n Updating list of available releases...
-n Done!
```

Is this related to the `echo -n`?

garazdawi:
I noticed the same thing and it is because on macOS `echo -n hello` prints `-n hello`. Apparently `-n` is a non POSIX extension to echo and therefore macOS `sh` has decided to not implement it. The solution is to use `printf "%s" "hello"` instead. I'm preparing a PR for that and some other issues that I found.

paulo-ferraz-oliveira:
Thanks, Lukas. You da man.

PR #410, opened 2022-05-01
Add autoclean on build failure

### LINKED ISSUE #262: kerl cleanup doesn't seem to work as described
I might be misunderstanding, but it looks like I should be able to use `kerl cleanup` on a specific version:

    $ ./kerl cleanup
    usage: ./kerl cleanup <build_name|all>

However when I run `kerl cleanup 20.1`, it looks like it removes the entire builds directory:

    $ ls -1 .kerl/builds/
    20.1
    20.2

    $ ./kerl cleanup 20.1
    Cleaning up compilation products for
    Cleaned up compilation products for  under /opt/erlang/.kerl/builds

    $ ls .kerl/builds
    ls: cannot access .kerl/builds: No such file or directory

    $ ls .kerl/
    archives  otp_builds  otp_installations  otp_releases

I'm using kerl 1.8.1 on CentOS 7.4

    $ ./kerl version
    1.8.1

    $ lsb_release -a
    LSB Version:    :core-4.1-amd64:core-4.1-noarch
    Distributor ID: CentOS
    Description:    CentOS Linux release 7.4.1708 (Core)
    Release:        7.4.1708
    Codename:       Core

Thanks!

crownedgrouse:
Configured by KERL_AUTOCLEAN variable (enabled by default).
Remove all build artefacts but logfile on failure if enabled.
This allow safe several attempts to build a release.

Logfile receive a banner for all build attempt with build variable at this time.

Fix #262

### DIFF
--- a/README.md
+++ b/README.md
@@ -256,6 +256,11 @@ Directory in which kerl will clone git repositories for building.
 
 ## Build configuration
 
+### KERL_AUTOCLEAN
+
+Default: 1 (Enabled)
+Clean all build artefacts but logfile on failure. This allow safe build retries after failure while still keeping logfile with all attempts logs, until success. Set to 0 to keep build artefacts on failure.
+
 ### KERL_CONFIGURE_OPTIONS
 
 Space-separated options to pass to `configure` when building OTP.
--- a/kerl
+++ b/kerl
@@ -644,10 +644,10 @@ show_configuration_warnings() {
 }
 
 show_logfile() {
-    echo "$1"
-    tail "$2"
-    echo
-    echo "Please see $2 for full details."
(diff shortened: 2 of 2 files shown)


jadeallenx:
We need to consolidate the steps for a failed build into a function like `_handle_build_failure()` which would:
* show the log file extract
* do an autoclean (if configured)
* exit with an appropriate exit status

This would also make it easy/cleaner to include new post failure cleanup in the future.

jadeallenx:
> Not sure to understand what you expect me to do, then.

Sorry, I didn't mean specifically for you to do this, just a clean up I'd like to do which is related to this PR.

PR #411, opened 2022-05-02
add KERL_RELEASE_TARGET

qzhuyan:


garazdawi:
Don't you also need to update the do_install so that beam.*.smp is copied over to the installed system?

Maybe it would be good to extend the Circle CI tests to make sure that this functionality works?

As an aside:

Since Erlang/OTP 22 it has been possible to do `make [install] TYPE=debug` at the ERL_TOP level and from Erlang/OTP 25 it should be possible to do the same with all the types described in this PR. So when kerl drops support for Erlang/OTP 24, it should be possible to make this a lot simpler.

Also, it is possible to use `erl -emu_type $TYPE` to start an emulator of a specific type. However, cerl is still useful for asan/valgrind as it starts Erlang using those tools with the correct options.

qzhuyan:
> Don't you also need to update the do_install so that beam.*.smp is copied over to the installed system?

Strange I only tested with 24.3.3 with  type `debug asan`, does not seem to need to copy beam.*.smp
```
$ which cerl
/home/xxxx/OTP/24.3.3-lttng/bin/cerl
$ find ~/OTP/24.3.3-lttng/ -name beam.*.smp
/home/xxxx/OTP/24.3.3-lttng/erts-12.3.1/bin/beam.debug.smp
$ cerl -asan
Erlang/OTP 24 [erts-12.3.1] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [jit] [address-sanitizer] [lttng]

Eshell V12.3.1  (abort with ^G)
1>

```

However, in build dir we have it.
```
$ find ~/.kerl/builds/24.3.3-lttng/ -name beam.*.smp
/home/xxxx/.kerl/builds/24.3.3-lttng/otp_src_24.3.3/bin/x86_64-pc-linux-gnu/beam.asan.smp
/home/xxxx/.kerl/builds/24.3.3-lttng/otp_src_24.3.3/bin/x86_64-pc-linux-gnu/beam.debug.smp
```



I will do the copy part in `do_install`.

crownedgrouse:
No check is done that values in variable is valid.
Something like this should be added : 
```
case $r_type in 
    opt|gcov|gprof|debug|valgrind|asan|lcnt)  
             # current code
        ;;
    *)
        l=w stderr "Invalid '$r_type' runtime type" 
esac
```

qzhuyan:
> No check is done that values in variable is valid. Something like this should be added :
> 
> ```
> case $r_type in 
>     opt|gcov|gprof|debug|valgrind|asan|lcnt)  
>              # current code
>         ;;
>     *)
>         l=w stderr "Invalid '$r_type' runtime type" 
> esac
> ```

Thanks for mention that, I will add it.
While testing  locally I find some runtime types are not supported in some OTP version.
Do you think we also need that check with Errors or Warnings?

kerl
```diff
@@ -1040,11 +1040,24 @@ _do_build() {
...
         l=n stderr "Also building Erlang/OTP $1 ($2) debug VM, please wait..."
         cd $ERL_TOP/erts/emulator || exit 1
         make debug ERL_TOP=$ERL_TOP >>"$LOGFILE" 2>&1
     fi
+
+    if [ -n "$KERL_BUILD_OTHER_RUNTIME_TYPES" ]; then
+    # where `$TYPE` is `opt`, `gcov`, `gprof`, `debug`, `valgrind`, `asan` or `lcnt`.
+        for r_type in $KERL_BUILD_OTHER_RUNTIME_TYPES;
+        do
+            echo "Also building Erlang/OTP $1 ($2) $r_type VM, please wait..."
```
qzhuyan, line 1056: will change it to `l=n stderr` style

crownedgrouse: for your question, I dunno.
I think kerl just have to check the global list of runtimes. Checking runtime type vs otp version is not necessary , and maybe complicated.
If a runtime type fails to be found, a warning could be neat, saying that this runtime time is maybe not available in this otp version.

qzhuyan: done. 

Print a warning if make runtime type is failed.

crownedgrouse:
Name of the variable is long to write and hard to remember.
You could keep it in code, but it could be initialized with KERL_BUILD_FLAVORS  (concatenated with KERL_BUILD_FLAVOURS... 😂 )
Your current variable should then be unset before.

qzhuyan:
renamed to `KERL_BUILD_FLAVOURS`

jadeallenx:
Would prefer `KERL_RELEASE_TARGET`

qzhuyan:
Just clarify about naming,  if we build  with `KERL_BUILD_FLAVOURS="valgrind debug"`

It will get a regular `erl` but also additional `cerl` for debugging 

After install and activate, We can start regular beam with

 `erl` 

or 
for debugging
`cerl -debug` or `cerl -valgrind`

Do you still want me to change to `KERL_RELEASE_TARGET`?

qzhuyan:
FYI, I added test to CI but don't know how to trigger it.

crownedgrouse:
maybe I missed things in code, but I do not see code using either FLAVORS or FLAVOURS.
Either both can be used, and be transparent, or impose to use one of both (dangerous).
I though that _KBF could be "$KERL_BUILD_FLAVORS $KERL_BUILD_FLAVOURS" in order to be safe.

For CI, I'm not owner of project, ask @jadeallenx ...

garazdawi:
> FYI, I added test to CI but don't know how to trigger it.

You can create an account on CircleCI and trigger CI runs on your fork as I have done here: https://app.circleci.com/pipelines/github/garazdawi/kerl

jadeallenx:
> Do you still want me to change to `KERL_RELEASE_TARGET`?

yes please; do not use a word with british and american spellings, we don’t need a repeat of behavior vs behaviour vs both 😅

jadeallenx:
Yes, thanks for this patch

Issue #413, opened 2022-06-06
Kerl Build Fails on Various MacOS Hosts and Various Releases


cogitaria-admin:
Here are a couple of failures just trying to get a bare-bones Erlang development environment setup. on MacOS  I have Xcode and updated openssl and such.  This is my Erlang version: Erlang/OTP 24 [erts-12.3.1] [source] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:1] [jit] [dtrace]

I'm running this build on macOS Monterey (also fails on Catalina but did work on that host in the past)
-------------------------
> kerl build 24.0 24.0
-n Extracting source code
-n Building Erlang/OTP 24.0 (24.0), please wait...
Build failed.
  const wxBitmap * Result = &This->GetBitmap();
                            ^~~~~~~~~~~~~~~~~~
----------------------------------------------------------
On the same box building 25.0 you get a different result:
--------------------
kerl build 25.0 25.0
-n Extracting source code
-n Building Erlang/OTP 25.0 (25.0), please wait...
Build failed.
gmake[4]: x86_64-apple-darwin21.4.0/Makefile: No such file or directory
gmake[4]: *** No rule to make target 'x86_64-apple-darwin21.4.0/Makefile'.  Stop.

I'm trying to "adopt Erlang" and this site was super awesome: https://adoptingerlang.org/ if only it worked (kerl build fails so you cannot get past **go**). 

Thanks for helping!

jadeallenx:
What version of kerl are you using? `kerl version` or `kerl status` should tell you.

garazdawi:
The problem with building Erlang/OTP 24.0 on macOS Monterey is because the WxWidget library did a breaking change in their API. The problem has been fixed with [this patch](https://github.com/erlang/otp/pull/5912) in [Erlang/OTP 24.3.4](https://www.erlang.org/patches/otp-24.3.4#OTP-18064). It might be a good idea (tm) for kerl to apply that patch to all Erlang versions.

Regarding the second problem, there should have been a message such as "Please see /some/file/path for full details.". Can you attach that log file to this issue?

cogitaria-admin:
Kerl version: 2.5.0

cogitaria-admin:
[otp_build_25.0.log](https://github.com/kerl/kerl/files/8863991/otp_build_25.0.log)

Attached the build log for the 25.0 build

garazdawi:
[...]

PR #414, opened 2022-06-09
Fix various issues

garazdawi:
This PR fixes various issues I found in the latest release of kerl.

1. Fix #409 to not use `echo -n` as that is not posix compliant (and actually breaks kerl on macOS)
2. Re-added shellcheck to CI and fixed all problems found. shellcheck would have found the bug with `echo -n`.
3. 885a31c37b678cb8a177f754ac51aa7e6f2dc9a9 inadvertently broke triggering an error when configure failed. Found when triaging #413.
4. Fix #409 to reset the terminal colors with sgr0 instead of color 9 as color 9 means different things in different terminals.
5. Fix #411 to fail and print logs if the build fails
6. Fix #411 to build debug build from ERL_TOP so that crypto and friends also are debug compiled
7. Added KERL_DEBUG for debugging kerl
8. Added testing for alpine linux packages when installing
9. Added Erlang/OTP 24 and 25 to CI tests
10. Fix the tests for #411 to work
11. Remove release of Erlang in build step, leftover from agner support.
12. Add log file for the install step

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -70,11 +70,11 @@ jobs:
               - run: |
                   apk update && apk upgrade
                   apk add curl ca-certificates \
-                    dpkg-dev dpkg \
                     gcc g++ libc-dev linux-headers make autoconf ncurses-dev tar \
                     openssl-dev lksctp-tools-dev lksctp-tools \
                     libxslt git
                   echo "export KERL_CONFIGURE_OPTIONS=--disable-hipe" >> $BASH_ENV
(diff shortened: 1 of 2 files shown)


jadeallenx (approved):

jadeallenx:
Thanks for this work! Very much appreciate it

garazdawi:
> Thanks for this work! Very much appreciate it

No problem! Thanks for maintaining kerl!

Would you be open to a PR migrating all the CI stuff to use github actions instead? I think it would help catch some faults as contributors do not need to create an additional account for it and we could also test on macOS.

PR #417, opened 2022-09-21
Make it a little less verbose

### LINKED ISSUE #416: kerl outputting result of uname | grep
I think this might be a recent addition to the code...

```
Extracting source code
Building Erlang/OTP 24.3.4.4 (24.3.4.4), please wait...
21.6.0
^ this
```

It confused me, since I was installing 24.3.4.4 and there was an extra output stating 21.6.0. It appears this output comes from https://github.com/kerl/kerl/blob/master/kerl#L858, which is potentially possible to hide. At this moment I just wanted to report this, so I don't forget it, and maybe in the future I might tackle it, if time allows and I don't forget 😄.

paulo-ferraz-oliveira:
Closes #416.

@garazdawi, your review would be appreciated. 🙇

### DIFF
--- a/kerl
+++ b/kerl
@@ -783,11 +783,7 @@ maybe_patch_catalina() {
 }
 
 is_osx() {
-    uname="$(uname -r | grep "^$1.*$")"
-    if [ "$uname" != "" ]; then
-        return
-    fi
-    false
+    uname -r | \grep "^$1.*$" > /dev/null
 }
 
 is_osx_catalina() {
--- a/kerl
+++ b/kerl
@@ -784,7 +784,9 @@ maybe_patch_catalina() {
 
 is_osx() {
     uname="$(uname -r | grep "^$1.*$")"
-    [ "$uname" != "" ] && return
+    if [ "$uname" != "" ]; then
+        return
+    fi
     false
 }
 
--- a/kerl
+++ b/kerl
@@ -782,8 +782,14 @@ maybe_patch_catalina() {
     fi
 }
 
+is_osx() {
+    uname="$(uname -r | grep "^$1.*$")"
+    [ "$uname" != "" ] && return
+    false
+}
+
 is_osx_catalina() {
-    uname -r | grep '^19.*$'
+    is_osx 19
 }
 
 compare_sem_version() {
@@ -851,11 +857,11 @@ maybe_patch_macos_version_check() {
 }
 
 is_osx_bigsur() {
-    uname -r | grep '^20.*$'
+    is_osx 20
 }
 
 is_osx_monterey() {
-    uname -r | grep '^21.*$'
+    is_osx 21
 }
 
 apply_bypass_version_check() {


paulo-ferraz-oliveira:
Is there a way to check the CircleCI results without having to create an account for that service?

kerl
```diff
@@ -782,8 +782,14 @@ maybe_patch_catalina() {
     fi
 }
 
+is_osx() {
+    uname="$(uname -r | grep "^$1.*$")"
+    [ "$uname" != "" ] && return
```
jadeallenx, line 787: I am not a huge fan of this construct - it's shorthand and cute but I'd prefer a more explicit and traditional if; then; else block, I guess.

paulo-ferraz-oliveira: Sure. No prob. 😄

paulo-ferraz-oliveira: How about 0f2ab1f? I may also change `return` to `return 0` and `false` to `return 1`, if you prefer.

kerl
```diff
@@ -782,8 +782,16 @@ maybe_patch_catalina() {
     fi
 }
 
+is_osx() {
+    uname="$(uname -r | grep "^$1.*$")"
+    if [ "$uname" != "" ]; then
+        return
+    fi
+    false
```
garazdawi, line 790:
```suggestion
is_osx() {
    uname -r | \grep "^$1.*$" > /dev/null
```

I would have written it like this. It is the form used in many other places in the `kerl` code.

paulo-ferraz-oliveira: Cool. I can change that. It's more compact. Thanks. @jadeallenx, are you Ok with this suggestion?

jadeallenx: Yes, definitely prefer that, thanks

jadeallenx:
> Is there a way to check the CircleCI results without having to create an account for that service?

I'm honestly pretty unhappy with CircleCI, no one maintains it and over time, and as usual it bit rots... I don't have a great solution for that.

paulo-ferraz-oliveira:
> > Is there a way to check the CircleCI results without having to create an account for that service?
> 
> I'm honestly pretty unhappy with CircleCI, no one maintains it and over time, and as usual it bit rots... I don't have a great solution for that.

Would you be interested in moving to GitHub Actions (I have some experience with this - haven't looked into the CircleCI stuff on this repo., to be honest)? We could start by creating an issue and maybe then implement it later...

kerl
paulo-ferraz-oliveira: Done. Thanks, @garazdawi, I just had to press the Commit suggestion button. 😄

jadeallenx:
GHA is fine by me

paulo-ferraz-oliveira:
btw, @jadeallenx, this is ready for review and merge. 😄

jadeallenx (approved):

jadeallenx:
Thanks!!

PR #427, opened 2022-11-30
Add CI for MacOS

paulo-ferraz-oliveira:
This adds MacOS-based CI with minimal changes to the workflow document.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,20 +17,30 @@ jobs:
       fail-fast: false
       matrix:
         otp_vsn: [23, 24, 25, master, latest]
-        os: [ubuntu-18.04, ubuntu-20.04]
+        os: [ubuntu-18.04, ubuntu-20.04, macos-11, macos-12]
     steps:
       - name: Update env.
         run: |
-          sudo sed -i 's/azure\.//' /etc/apt/sources.list # Reduces chance of time-outs
-          sudo apt-get update -y
-          sudo apt-get upgrade -y
-          sudo apt-get install -y --no-install-recommends \
-            curl ca-certificates git autoconf dpkg-dev \
-            zlib1g-dev libncurses5-dev libssl-dev unixodbc-dev libgmp3-dev \
-            libwxgtk3.0-gtk3-dev libwxgtk-webview3.0-gtk3-dev libsctp-dev \
-            lksctp-tools build-essential gcc-9 m4 fop xsltproc \
-            default-jdk libxml2-utils procps valgrind binutils
-          echo 'KERL_RELEASE_TARGET=debug opt gcov gprof valgrind lcnt' >> $GITHUB_ENV
+          if [[ ${{matrix.os}} == macos* ]]; then
+            /bin/bash -c "$(curl -fsSL \
+              https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
+            brew install automake fop wxwidgets
+          else
+            sudo sed -i 's/azure\.//' /etc/apt/sources.list # Reduces chance of time-outs
+            sudo apt-get update -y
+            sudo apt-get upgrade -y
+            sudo apt-get install -y --no-install-recommends \
+              curl ca-certificates git autoconf dpkg-dev \
+              zlib1g-dev libncurses5-dev libssl-dev unixodbc-dev libgmp3-dev \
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -17,20 +17,30 @@ jobs:
       fail-fast: false
```
paulo-ferraz-oliveira, line 17: Changes are probable best viewed with "Hide whitespace", since this introduces an `if` which changes existing script indentation.

.github/workflows/ci.yml
```diff
@@ -17,20 +17,30 @@ jobs:
...
-          sudo apt-get upgrade -y
-          sudo apt-get install -y --no-install-recommends \
-            curl ca-certificates git autoconf dpkg-dev \
-            zlib1g-dev libncurses5-dev libssl-dev unixodbc-dev libgmp3-dev \
-            libwxgtk3.0-gtk3-dev libwxgtk-webview3.0-gtk3-dev libsctp-dev \
-            lksctp-tools build-essential gcc-9 m4 fop xsltproc \
-            default-jdk libxml2-utils procps valgrind binutils
-          echo 'KERL_RELEASE_TARGET=debug opt gcov gprof valgrind lcnt' >> $GITHUB_ENV
+          if [[ ${{matrix.os}} == macos* ]]; then
+            /bin/bash -c "$(curl -fsSL \
```
paulo-ferraz-oliveira, line 25: The only differences, from Ubuntu, are this section and the one below (`KERL_RELEASE_TARGET`). I use Homebrew to ease the installation process of required tools...

.github/workflows/ci.yml
```diff
@@ -17,20 +17,30 @@ jobs:
...
+            sudo apt-get upgrade -y
+            sudo apt-get install -y --no-install-recommends \
+              curl ca-certificates git autoconf dpkg-dev \
+              zlib1g-dev libncurses5-dev libssl-dev unixodbc-dev libgmp3-dev \
+              libwxgtk3.0-gtk3-dev libwxgtk-webview3.0-gtk3-dev libsctp-dev \
+              lksctp-tools build-essential gcc-9 m4 fop xsltproc \
+              default-jdk libxml2-utils procps valgrind binutils
+          fi
+          if [[ ${{matrix.os}} == macos* ]]; then
+            echo 'KERL_RELEASE_TARGET=debug opt' >> $GITHUB_ENV
```
paulo-ferraz-oliveira, line 40: I single this `if` out so it's easier to check for the differences (and hopefully one day they can be the same - btw, we're not doing target `asan` for Ubuntu).

The targets were not testing aren't straightforward to achieve, at least with Homebrew. e.g. `valgrind` would come from [an alternative source](https://github.com/LouisBrunner/homebrew-valgrind), `gcov` and `gprof` supposedly comes from [Command Line Tools](https://mac.install.guide/commandlinetools/index.html), but I can't seem to find `gcov` after installing those, and `gprof`'s been deprecated, ...

If all goes well this can still be an acceptable first base for building upon.

jadeallenx (approved):

paulo-ferraz-oliveira:
@jadeallenx, ready to merge when you are 😄

jadeallenx:
Thanks

jadeallenx:
At this point I want to give you a commit bit for this repo if you want one...

.github/workflows/ci.yml
garazdawi: If `debug` and `opt` work then the rest should also work. No need to make things extra complicated in macOS.

paulo-ferraz-oliveira: Hey, @garazdawi. I tried it with `debug opt gcov`, for example, and stuff started failing (thus the `if`). I didn't want to dig further, since I believed this was Ok for a first step, but maybe I could open a pull request where I add just `gcov` (so you can see the issue) and maybe have your 👀.

paulo-ferraz-oliveira:
Hey, @jadeallenx. What's a commit bit? I don't think I've come across the term. Is that like collaboration rights? I'd be Ok with that, sure. Thank you.

jadeallenx:
> Hey, @jadeallenx. What's a commit bit? I don't think I've come across the term. Is that like collaboration rights? I'd be Ok with that, sure. Thank you.

It's permission to make commits to this repository

Issue #402, opened 2022-03-21
'error: process ID out of range' when running an container


winks:
I am trying to install erlang 24.3.2 in a redhat/ubi8 container and kerl errors out in the `SHELL_SUFFIX` check.

```
Removing intermediate container 0f8337abdd79
 ---> 335f5f14b7fc
Step 8/10 : RUN 	cd /tmp && ./kerl install 24.3.2 /opt/erlang/24.3.2
 ---> Running in 9d2d6f3f2a02
Installing Erlang/OTP 24.3.2 (24.3.2) in /opt/erlang/24.3.2...
error: process ID out of range

Usage:
 ps [options]

 Try 'ps --help <simple|list|output|threads|misc|all>'
  or 'ps --help <s|l|o|t|m|a>'
 for additional help text.

For more details see ps(1)
```

This is the Dockerfile I use at the moment (just started, that's why there are many layers)

```
# docker build -t foo .
FROM redhat/ubi8

RUN \
    yum install -y \
        autoconf \
        automake \
        gcc \
        git \
        make \
        ncurses-devel \
        openssl-devel

RUN \
    cd /tmp && \
    curl -O https://raw.githubusercontent.com/kerl/kerl/master/kerl && \
    chmod a+x kerl && \
    ./kerl build 24.3.2

RUN \
    yum install -y \
        unixODBC \
        unixODBC-devel

RUN \
    yum install -y \
        procps

RUN \
    cd /tmp && ./kerl install 24.3.2 /opt/erlang/24.3.2
````

This is how it looks interactively:

```
[root@c071e2ea1c6a /]# echo $$
62
[root@c071e2ea1c6a /]# export PID=$$; echo $PID
62
[root@c071e2ea1c6a /]# export PARENT_PID=$(ps -p $PID -o ppid=)
[root@c071e2ea1c6a /]# echo $PARENT_PID 
1
[root@c071e2ea1c6a /]# export PARENT_CMD=$(ps -p $PARENT_PID -o ucomm | tail -n 1)
[root@c071e2ea1c6a /]# echo $PARENT_CMD
bash

```

And this is what seems to happen:

```
[root@c071e2ea1c6a /]# export PID=$$; echo $PID
1
[root@c071e2ea1c6a /]# export PARENT_PID=$(ps -p $PID -o ppid=)
[root@c071e2ea1c6a /]# echo $PARENT_PID
0
[...]

PR #428, opened 2022-11-30
Fix error when running in container (#402)

winks:
Add a check that looking up the parent PID of process 0 doesn't fail


jadeallenx (approved):

jadeallenx:
Will merge, pending CI runs

jadeallenx:
Thank you so much!

PR #429, opened 2022-12-01
Test on Ubuntu 22.04 (while dropping 20.04)

paulo-ferraz-oliveira:
This PR introduces Ubuntu 22.04 and drops (deprecated Ubuntu 18.04), while also addressing some previous comments by @garazdawi, namely:

- dropping `latest` (in favor of `25` - already present in the CI matrix)
- decreasing the number of release targets (for faster CI) - if I got @garazdawi's comment right 😓 (**edit**: it seems to shave off at least around 20 min., from a previous 60+ min.)
- moving `$(getconf _NPROCESSORS_ONLN)` to `$(nproc)`, for readability

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,7 +17,7 @@ jobs:
       fail-fast: false
       matrix:
         otp_vsn: [23, 24, 25, master, latest]
-        os: [ubuntu-18.04, ubuntu-20.04, macos-11, macos-12]
+        os: [ubuntu-20.04, ubuntu-22.04, macos-11, macos-12]
     steps:
       - name: Update env.
         run: |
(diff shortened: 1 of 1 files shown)


paulo-ferraz-oliveira:
Ubuntu 22.04 / OTP 23.3.4.18 ❌: https://github.com/kerl/kerl/actions/runs/3596929436/jobs/6058147491#step:6:5316 

@garazdawi, is this issue know to you?

paulo-ferraz-oliveira:
Also, should we care about:

- OpenSSL 3.x -specific warning https://github.com/kerl/kerl/actions/runs/3596929436/jobs/6058147621#step:6:1009 (seems to be a 25- issue, only)
- wxWidgets -specific warning https://github.com/kerl/kerl/actions/runs/3596929436/jobs/6058147621#step:6:1009 (seems to affect MacOS builds, only - and when digging up via Homebrew it seems the expected flag, for previous versions `...compat28` is not set - then there's various workarounds for this that usually involve manually updating the brew recipes, something I'd prefer to avoid doing)

PR #433, opened 2023-01-27
Include MacOS Ventura in bypass version check

jjcarstens:
Attempting to install older OTP versions on MacOS Ventura will fail because the version check bypass does not get applied.

This adds the simple check for the newest version of MacOS to include it in the version check bypass

Tested with OTP 23.0.4


paulo-ferraz-oliveira:
The `Lint` workflow issue should be related to https://github.com/kerl/kerl/pull/438. I believe this change is ready to be merged. @jadeallenx, thoughts?

Issue #434, opened 2023-01-30
Building Erlang 23 freezes on macOS 


eproxus:
If I try to build 23.3.4.18 on macOS Ventura 13.2 it freezes when patching `configure.in` at this point: https://github.com/kerl/kerl/blob/master/kerl#L768

If I check the build log at `$KERL_HOME/builds/my_build_23.3.4.18/otp_build_23.3.4.18.log` it only shows this text:

```
patching file 'erts/configure.in'
Reversed (or previously applied) patch detected!  Assume -R? [y] 
```

As far as I can tell, that patch is already in OTP 23 (see the list of tags here https://github.com/erlang/otp/commit/4b0467c).

In fact, it even seems that the configure scripts are checked in for OTP 23 so maybe `KERL_USE_AUTOCONF=1` shouldn't be used at all?

paulo-ferraz-oliveira:
Hey.

Yeah, I though about that in the past (that the configure scripts are checked in, and the option might be deprecated by this time around). Care to pull request removing the interesting bits 😄? CI should be able to tell us if stuff's as it should.

Cheers.

eproxus:
If I look at the OTP commit it seems it is in 23.1+, but not e.g. 23.0. I guess there's no easy way to do that test for minor versions in kerl?

paulo-ferraz-oliveira:
There's something similar [here](https://github.com/kerl/kerl/blob/master/kerl#L781). Does that help?

electricshaman:
Hilariously, I just noticed I can get past the freeze by hitting enter at the prompt regarding the patch.  It's waiting for input, not doing anything.  Not ideal for an automated script, but at least the build moves forward.

PR #437, opened 2023-04-05
Ignore already applied patch

magicant:
Fixes https://github.com/kerl/kerl/issues/434

### DIFF
--- a/kerl
+++ b/kerl
@@ -810,7 +810,7 @@ compare_sem_version() {
 }
 
 apply_catalina_no_weak_imports_patch() {
-    patch -p1 <<'_END_PATCH'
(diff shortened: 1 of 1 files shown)


paulo-ferraz-oliveira:
This should be Ok to merge (the CI issue should be fixed by https://github.com/kerl/kerl/pull/438), but I'm wondering if we could apply the same fix to all those other places where we use `patch` (?) @jadeallenx, thoughts?

jadeallenx:
I think we probably should using `-N` elsewhere yes.

magicant:
Thanks for the suggestion, I've updated the patch. @paulo-ferraz-oliveira @jadeallenx

PR #443, opened 2023-04-20
Force OpenSSL 1.1.1f on Ubuntu 22.04, for CI purposes

paulo-ferraz-oliveira:
Fixes #440.

PR #444, opened 2023-05-04
Install OpenSSL 1.1 in a less generic manner

paulo-ferraz-oliveira:
Hopes to address https://github.com/kerl/kerl/pull/443#pullrequestreview-1394874612.


paulo-ferraz-oliveira:
@jadeallenx, this update tries to be smarter on obtaining files.

We use filters, like e.g. `openssl_1.1*`.
We download all files that respect those filters.
We sort these and get the most recent ones (`openssl`, `libssl`, and `libssl-dev`).
We install those.

This should bypass the issue we're having at the moment, but maybe it doesn't address your genericness concern (?) Lemme know.

Oh, and if test pass and you wanna you can merge, because this is basically affecting all pull requests on that specific workflow matrix combo.

PR #445, opened 2023-05-10
Remove docsh from kerl

jadeallenx:
Since a `chunks` format now exists to pull in Erlang docs to the REPL docsh is no longer needed for supported OTP builds.


jadeallenx:
I want to remove docsh and then do a new release - probably will tag it 3.0

paulo-ferraz-oliveira (approved):
LGTM

paulo-ferraz-oliveira:
3.0 should be Ok, since it breaks backward compatibility. 👍

PR #446, opened 2023-06-26
Respect markdownlint

paulo-ferraz-oliveira:
A very minor thing but enough to have 
image
in the frontpage 😄


jadeallenx (approved):

jadeallenx:
yeah that was my bad- sorry

PR #448, opened 2023-07-04
Modernize macos on CI

paulo-ferraz-oliveira:


paulo-ferraz-oliveira:
Closing this in favor or https://github.com/kerl/kerl/pull/449, which also does the modernisation mentioned herein.

PR #452, opened 2023-08-13
Add command emit-activate (sh, bash, fish, csh)

### LINKED ISSUE #451: Add 'kerl build source' to build from source
When I'm hacking on Erlang/OTP, I like to have `use erlang source` in my `.envrc`; see https://blog.differentpla.net/blog/2019/01/30/integrating-kerl-direnv/ (though it's not strictly relevant to this issue).

At the moment, I have to "persuade" kerl to use my built-from-source; see https://blog.differentpla.net/blog/2023/01/10/erlang-source-with-direnv/

It would be nice if kerl had better support for building and (faking?) install from source.

Ideally, I'd just `kerl build source ~/Source/rlipscombe/otp source` and it would build in-place, and then `kerl install source ~/erlangs` would do the right thing.

paulo-ferraz-oliveira:
# Description

As per the updated README...

### `emit-activate`

```console
$ kerl emit-activate <release> <build_name> <directory> [sh|bash|fish|csh]
```

Print the `activate` script, as generated by `kerl`.

**Note**: the output defaults to shell `sh` (also `bash`), as per file `activate`. To emit
`activate.fish`, or `activate.csh`, set the last argument to `fish` or `csh`, respectively.

Closes #451.

**Edit**: while hacking on #454 I found that we could probably benefit the activation scripts by including a `#!` on top of the ones that don't have it:

* the bash one works as is with `./activate` (it has a shebang),
* the fish one emits a warnings with `./activate` (it has no shebang), and
* the csh one has text explaining you should use `source`.

For this pull request I didn't have to install shells or test the code, but simply compare the emitted (from the new command) to the existing (from older installs) whereas for the new pull request I had to install the shells to test the differences.


.github/workflows/ci.yml
```diff
@@ -47,6 +41,7 @@ jobs:
               echo '_KERL_PREFIX_GIT=git' >> $GITHUB_ENV
               echo '_KERL_PREFIX_GIT_TARGET=https://github.com/erlang/otp.git' >> $GITHUB_ENV
               echo 'KERL_BUILD_DOCS=true' >> $GITHUB_ENV
+              _VERSION=git
```
paulo-ferraz-oliveira, line 44: Without this there's be a mismatch on the generated scripts, since internally we use `git` for branch-based builds, not the branch itself.

paulo-ferraz-oliveira:
There's still some minor CI-based issues I want to fix... I'll request a review when I'm ready.

paulo-ferraz-oliveira:
@jadeallenx, I asked @rlipscombe to test this out (in the linked issue), but I believe it can already be reviewed.

jadeallenx (approved):

jadeallenx:
I'd like to add some kind of environment variable that controls _what_ shells to emit for (default to current behavior for all shells) and the _name_ of the file to emit into. This might be useful for folks who use fish and want their activate script to be called `activate` and not `activate.fish` or w/e.

Anyway, random thought. Probably will open an issue for it. This is great, thank you <3

Issue #458, opened 2023-09-10
Improve activate/deactivate scripts


x0id:
1. The deactivation script should take into account if the variable of interest was set before. If it was not set, it should be unset after de-activation
2. Same logic to be applied to MANPATH with minor addition - if it was not set before, _and is empty after deactivation is done_, unset it.
3. No extra variables should be left after deactivation (except for MANPATH if it was manipulated between activate and deactivate steps).
4. These rules should be covered by the test cases for sh/fish/csh.

PR #460, opened 2023-09-17
Fix fish shell (de)activate

x0id:
# Description

The fish shell kerl activate/deactivate code has been improved and refactored to fix environment setup/cleanup issues (#458).

The test suite for fish shell based activate/deactivate has been added.

This PR partially (for the fish shell) resolves issue #458.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,7 +18,7 @@ jobs:
           if [[ ${{matrix.os}} == macos* ]]; then
             /bin/bash -c "$(curl -fsSL \
               https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
-            brew install automake fop wxwidgets
+            brew install automake fop wxwidgets fish
           else
             sudo sed -i 's/azure\.//' /etc/apt/sources.list # Reduces chance of time-outs
             sudo apt-get update -y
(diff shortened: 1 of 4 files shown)


paulo-ferraz-oliveira:
👍 very nice, @x0id, thanks.

I don't know enough `fish` to do a proper review on this, but it seems solid, especially since it's tested (if I understood correctly, similar to its `bash` counterpart).

@jadeallenx, I'm Ok to merge as is and then update/fix if there's complaints from the community. What do you say?

paulo-ferraz-oliveira:
And, @x0id, if you think you'll get around to the rest of #458 soonish, we can probably postpone a release until then, otherwise we can release a patch now and patch later. Thoughts?

x0id:
@paulo-ferraz-oliveira, I think there is no rush in releasing the fish shell updates, I'm not sure actually how many people use it. I plan to work on sh/csh in the next couple of weeks.

jadeallenx:
> @jadeallenx, I'm Ok to merge as is and then update/fix if there's complaints from the community. What do you say?

I think we should merge it too.

jadeallenx:
> I'm not sure actually how many people use it. I plan to work on sh/csh in the next couple of weeks.

Nothing makes those hidden users show up than a new release that breaks something lol 😄

jadeallenx (approved):

paulo-ferraz-oliveira:
It's squash-merged. Thanks all.

PR #465, opened 2023-09-26
Trim `kerl list releases`

### LINKED ISSUE #459: Trim `list releases` to supported versions

A visual one, maybe...

Since our maintenance policy is basically the same as OTP's, I propose we trim the list of "list releases" to supported versions alone. This can be a variable in the `kerl` script itself, that we'll only have to bump around once per year.

None.

#### Additional context

None.

paulo-ferraz-oliveira:
# Description

... as per the `README`'s Erlang/OTP support policy, and the discussion below.

Closes #459.

### DIFF
--- a/kerl
+++ b/kerl
@@ -134,6 +134,7 @@ KERL_BUILD_DIR=${KERL_BUILD_DIR:="${KERL_BASE_DIR:?}"/builds}
 KERL_GIT_DIR=${KERL_GIT_DIR:="${KERL_BASE_DIR:?}"/gits}
 KERL_GIT_BASE=https://raw.githubusercontent.com/kerl/kerl/master
 KERL_COLORIZE=${KERL_COLORIZE:=$KERL_COLOR_AVAILABLE}
+KERL_INCLUDE_RELEASE_CANDIDATES=${KERL_INCLUDE_RELEASE_CANDIDATES:=no}
 
 if [ -n "$OTP_GITHUB_URL" ]; then
     _OGU="$OTP_GITHUB_URL"
@@ -1933,29 +1934,43 @@ list_print() {
             if [ "${list_type}" = releases ] && [ "$maybe_all" != all ]; then
                 awk -v oldest_listed="$OLDEST_OTP_LISTED" \
                     -v oldest_supported="$OLDEST_OTP_SUPPORTED" \
-                    -v last_major="$OLDEST_OTP_LISTED" '
+                    -v include_rc="$KERL_INCLUDE_RELEASE_CANDIDATES" '
+                function _print(vsn, is_this_supported) {
+                    if (is_this_supported) {
+                        suf=" *"
+                    }
+                    printf "%s%s\n", vsn, suf
+                }
                 {
-                    full_version=$0
-                    split(full_version, version_components, ".")
-                    major=version_components[1]
-                    if (major >= oldest_listed && !/-rc/ && !/^R/) {
-                        if (major > last_major) {
-                            printf "%s",last_full_version
+                    this_version=$0
+                    split(this_version, version_components, ".")
+                    this_major=version_components[1]
 
-                            if (last_major >= oldest_supported) {
-                                printf " *"
-                            }
+                    if (last_major == "") {
+                        last_major = oldest_supported - 1
+                    }
 
-                            printf "\n"
-                        }
-                        last_major=major
-                        last_full_version=full_version
+                    is_this_r=/^R/
+                    is_this_rc=/-rc/
+                    is_rc_printable=(include_rc == "yes")
+                    is_this_supported=(last_major >= oldest_supported)
+
+                    if (!is_this_r) {
+                        is_transition_to_rc=(is_last_not_rc && is_this_rc)
+                        is_transition_from_rc=(!is_last_not_rc && !is_this_rc && is_rc_printable)
+                        (is_transition_to_rc || is_transition_from_rc) \
+                            && _print(last_version, is_this_supported)
+
+                        last_version=this_version
+                        last_major=this_major
+                        is_last_not_rc=!is_this_rc
                     }
                 }
                 END {
-                    if (!/-rc/) {
-                        printf "%s *\n",$0
-                    }
+                    last_version=$0
+                    is_supported=1
+                    (!is_this_rc || is_rc_printable) \
+                        && _print(last_version, is_supported)
                 }' "$list"
             else
                 cat "$KERL_BASE_DIR/otp_$1"
--- a/kerl
+++ b/kerl
@@ -41,6 +41,7 @@ if [ -n "$KERL_DEBUG" ]; then
 fi
 
 KERL_VERSION='3.1.0'
+OLDEST_OTP_LISTED='17'
 OLDEST_OTP_SUPPORTED='24'
 
 ERLANG_DOWNLOAD_URL='https://erlang.org/download'
@@ -1924,15 +1925,34 @@ maybe_remove() {
 
 list_print() {
     list_type=$1 # releases | builds | installations
+    maybe_all=$2
+
     list=$KERL_BASE_DIR/otp_${list_type}
     if [ -f "$list" ]; then
         if [ "$(\wc -l "$list")" != '0' ]; then
-            if [ "${list_type}" = releases ]; then
-                awk -v target="$OLDEST_OTP_SUPPORTED" '
+            if [ "${list_type}" = releases ] && [ "$maybe_all" != all ]; then
+                awk -v oldest_listed="$OLDEST_OTP_LISTED" \
+                    -v oldest_supported="$OLDEST_OTP_SUPPORTED" \
+                    -v last_major="$OLDEST_OTP_LISTED" '
                 {
-                    split($0, v, ".")
-                    if (v[1] >= target && !/^R/) {
-                        printf "%s\n",$0
+                    full_version=$0
+                    split(full_version, version_components, ".")
+                    major=version_components[1]
+                    if (major >= oldest_listed && !/-rc/ && !/^R/) {
+                        if (major > last_major) {
+                            if (last_major >= oldest_supported) {
+                                printf "* "
+                            }
+
+                            printf "%s\n",last_full_version
+                        }
+                        last_major=major
+                        last_full_version=full_version
+                    }
+                }
+                END {
+                    if (!/-rc/) {
+                        printf "* %s\n",$0
                     }
                 }' "$list"
             else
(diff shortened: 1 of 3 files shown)


kerl
```diff
@@ -1922,9 +1923,22 @@ maybe_remove() {
...
+                        printf "%s\n",$0
+                    }
+                }' "$list"
+            else
+                cat "$KERL_BASE_DIR/otp_$1"
+            fi
+
             return 0
         fi
     fi
```
paulo-ferraz-oliveira, line 1944: The goal is not to change internal behaviour (or the actual `releases` file) but rather just trim the list. A further pull request will remove dead code, effectively changing "the support".

In any case do lemme know if you think the generated `otp_releases` should also be trimmed (e.g. for consistency), in which case this code would be reverted but the code to create the file would be updated.

kerl
```diff
@@ -1922,9 +1923,22 @@ maybe_remove() {
...
+            if [ "${list_type}" = releases ]; then
+                awk -v target="$OLDEST_OTP_SUPPORTED" '
+                {
+                    split($0, v, ".")
+                    if (v[1] >= target && !/^R/) {
+                        printf "%s\n",$0
+                    }
+                }' "$list"
+            else
+                cat "$KERL_BASE_DIR/otp_$1"
```
paulo-ferraz-oliveira, line 1939: Kept as-is for anything other than releases (e.g. builds and installations).

jadeallenx:
I don't think we should artificially limit the list of valid releases but it think it _would_ be helpful to mark which OTP releases we are "officially" supporting somehow - maybe with an asterisk like git branch?  Also the list is ridiculously long now - it might be worthwhile to trim some of it that's really quite ancient by default - if you give like "all" as a parameter you get the entire list, otherwise, you get a shorter more recently set of releases by default

paulo-ferraz-oliveira:
Ok, so based on your comment:

1. we could accept `kerl list releases all` to have a full list (basically as-is, today, with a potential filter on top - see 2.)
2. we could change `kerl list releases` to "a shorter more recently set of releases by default" - do you want a fix "initial version", e.g. latest - 4?
3. accept `kerl list releases` with an `*` for the ones that are in the support policy (and a note, at the bottom stating what the `*` is)

In any case, do you think `otp_releases` should always contain all versions, even if we're filtering some for "listing"?

jadeallenx:
> Ok, so based on your comment:
> 
> 1. we could accept `kerl list releases all` to have a full list (basically as-is, today, with a potential filter on top - see 2.)
> 2. we could change `kerl list releases` to "a shorter more recently set of releases by default" - do you want a fix "initial version", e.g. latest - 4?
> 3. accept `kerl list releases` with an `*` for the ones that are in the support policy (and a note, at the bottom stating what the `*` is)

This all sounds excellent

> In any case, do you think `otp_releases` should always contain all versions, even if we're filtering some for "listing"?

I do think it should be the entire list - I wish there was a way we could not fetch it so often but afaik, there's no HEAD method we could use

garazdawi:
I have some scripts that parse the output of `kerl list releases` so if you can it would be great to not change the output too much :)

Maybe it could:
* remove any release before 17 (those starting with R)
* remove any release candidate
* list only the latest patch for each release

For example:

```
> kerl list releases
17.5.6.10
18.3.4.11
19.3.6.13
20.3.8.26
21.3.8.24
22.3.4.26
23.3.4.19
24.3.4.13
25.3.2.6
26.1
Run 'kerl update releases' to update this list from erlang.org
Run 'kerl list releases all' to view all available releases
```

jadeallenx:
I think @garazdawi is not alone in having some tools that parse kerl's output so I think we need to continue to provide the entire list but I think these suggestions for pruning are helpful and good ones.  The pruned list probably should include some text like "the full list of release is available using the command ..."

paulo-ferraz-oliveira:
I understand the need to have output similar to what there is, right now, but that could potentially be controlled with a variable, which is, I guess, the purpose of `all` (doesn't that suit you, @garazdawi?). So we'd trim the default but allow keeping behaviour via a control flag. We could then bump the major version and specify why that is, e.g. in the release notes.

paulo-ferraz-oliveira:
Ok, so what I have until now is:

- we make (new) `kerl list releases all` behave the same as (atm) `kerl list releases`
- we (breaking change) update `kerl list releases` to "a shorter more recently set of releases by default" with a fixed "initial version"
  - e.g. latest - 9 (this is an implementation detail we can discuss in the review pane)?
    - what I don't like about this approach is the support policy is OTP-2 but we're printing stuff that's e.g. 17+ (thus the `- 9`)
  - in this list we also remove release candidates, and show only latest patches
  - we add, to `kerl list releases`, an `*` for the ones that are in the support policy (and a note, at the bottom stating what the `*` is), but only for the regular list (again, `all` keeps behaviour, as stated)
- `otp_releases` contain all versions, as-is atm
- we add note "Run 'kerl list releases all' to view all available releases" to the bottom of the non-`all` call

Does this seem to suit all identified use cases, @jadeallenx, @garazdawi?

jadeallenx:
This is a great summary! Yes let’s do this please

paulo-ferraz-oliveira:
@garazdawi, I'm moving in the direction of my previous comment (with @jadeallenx's blessing 😄) . Do let me know if there's something you think could be adjusted, otherwise I can also add you as a reviewer and then we can go over the final result together.

paulo-ferraz-oliveira:
Output for `kerl list releases all` (culled below):

```console
R13B03
...
R16B03_yielding_binary_to_term
17.0-rc1
...
18.0
...
19.0.1
...
20.1
...
21.2.1
...
22.3.4.26
...
26.1
Run './kerl update releases' to update this list from erlang.org
```

Output for `kerl list releases`:

```console
17.5.6.10
18.3.4.11
19.3.6.13
20.3.8.26
21.3.8.24
22.3.4.26
23.3.4.19
* 24.3.4.13
* 25.3.2.6
* 26.1
Run './kerl update releases' to update this list from erlang.org
Run './kerl list releases all' to view all available releases
Note: * means "currently supported"
```

Given the current implementation, this also means that e.g. when `27.0-rc1` comes along it won't be visible in the non-`all` list.

garazdawi:
Can we have the star after the version so that it is an easy grep + awk to get the versions?

`grep “^[0-9]” | awk '{print $1}'`

kerl
```diff
@@ -2526,20 +2560,36 @@ upgrade)
...
+            exit 1
+        fi
+        if [ $# -eq 3 ] && [ "$3" != 'all' ]; then
+            list_usage
+            exit 1
+        fi
         check_releases
-        list_print "$2"
+        list_print "$2" "$3"
         l=t stderr "Run '$0 update releases' to update this list from erlang.org"
```
jadeallenx, line 2575: Since we're touching this area of code, I think this message should not mention the erlang.org website any more (since it really parses tags from github by default)

```suggestion
        l=t stderr "Run '$0 update releases' to update this list"
```

kerl
```diff
@@ -1922,9 +1924,41 @@ maybe_remove() {
...
+
+                            printf "%s\n",last_full_version
+                        }
+                        last_major=major
+                        last_full_version=full_version
+                    }
+                }
+                END {
+                    if (!/-rc/) {
+                        printf "* %s\n",$0
```
jadeallenx, line 1955: Put the asterisk behind the version tag
```suggestion
                        printf "%s *\n",$0
```

jadeallenx (commented):
This all looks good to me. I think maybe we should only list releases from 20 onwards personally. But I'm willing to hear reasons why we should include older releases (17 has barely baked map support, for example...)

jadeallenx:
I feel like there should be a flag like "KERL_INCLUDE_RELEASE_CANDIDATES=yes" if you want rcs to be listed in your o/p?

kerl
paulo-ferraz-oliveira: I'll accept the suggestion, but also change the README at the same time, so I can't commit from this one 😄

paulo-ferraz-oliveira:
I'll add flag `KERL_INCLUDE_RELEASE_CANDIDATES`.

kerl
paulo-ferraz-oliveira: 5ae4131

kerl
paulo-ferraz-oliveira: 5ae4131

paulo-ferraz-oliveira:
As for the minimum versions (listed or supported) feel free to suggest changes to the variables that define those (instead of `17` it could be `20` or something else). It's a minor detail, to me.

paulo-ferraz-oliveira:
Here's some output (with release candidates)

```console
$ KERL_INCLUDE_RELEASE_CANDIDATES=yes ./kerl list releases
17.0-rc2
17.5.6.10
18.0-rc2
18.3.4.11
19.0-rc2
19.3.6.13
20.0-rc2
20.3.8.26
21.0-rc2
21.3.8.24
22.0-rc3
22.3.4.26
23.0-rc3
23.3.4.19
24.0-rc3 *
24.3.4.13 *
25.0-rc3 *
25.3.2.6 *
26.0-rc3 *
26.1.1 *
Run './kerl update releases' to update this list
Run './kerl list releases all' to view all available releases
Note: * means "currently supported"
```

paulo-ferraz-oliveira:
Here's some output (without release candidates)

```console
$ ./kerl list releases
17.5.6.10
18.3.4.11
19.3.6.13
20.3.8.26
21.3.8.24
22.3.4.26
23.3.4.19
24.3.4.13 *
25.3.2.6 *
26.1.1 *
Run './kerl update releases' to update this list
Run './kerl list releases all' to view all available releases
Note: * means "currently supported"
```

paulo-ferraz-oliveira:
Whatever the values that end up in those control variables (for older versions) I feel this is already a major improvement for what we have currently.

jadeallenx:
> Whatever the values that end up in those control variables (for older versions) I feel this is already a major improvement for what we have currently.

Yes, no question - this is a lovely improvement. Thanks!

jadeallenx (approved):

jadeallenx:
We can leave at 17 for now I guess.

paulo-ferraz-oliveira:
Perfect stuff. @garazdawi, feel free to still comment on the final result, even if merged, as I might've not addressed all your concerns. In any case, I'm gonna move forward with other stuff I have prepared...

garazdawi:
lgtm!

paulo-ferraz-oliveira:
And as soon as it was merged it bit me 😄 I started a repo. with @starbelly that pulls `kerl` from the main branch, and we were using `list releases` hahaha (now it's using `list releases all`).

PR #466, opened 2023-09-28
Bury dead code

### LINKED ISSUE #455: Remove no-longer-supported OTP version-specific code
Remove pre-24 -specific code, as per the [Erlang/OTP support policy](https://github.com/kerl/kerl#erlangotp-support-policy).

If 27, 28, etc. is out before this is handled pre-24 will become pre-25, pre-26, etc.

Examples of this:

* [maybe_patch_darwin](https://github.com/kerl/kerl/blob/17943479c00e53b5e17102119d4c371c55d737bc/kerl#L777C1-L777C19)
* [maybe_patch_catalina](https://github.com/kerl/kerl/blob/17943479c00e53b5e17102119d4c371c55d737bc/kerl#L792C1-L792C21)

paulo-ferraz-oliveira:
# Description

Closes #455.


kerl
```diff
@@ -953,10 +734,10 @@ _flags() {
             CC='clang'
         fi
 
-        CFLAGS="$CFLAGS" CC="$CC" "$@"
+        CFLAGS="${CFLAGS:-}" CC="$CC" "$@"
```
paulo-ferraz-oliveira, line 737: I couldn't find these being used, after my changes, but ShellCheck complains about them, so I'm assuming they might come from the consumer environment, and just set them to default "".

paulo-ferraz-oliveira:
Just look at it go... 👀

kerl
```diff
@@ -724,194 +693,6 @@ show_logfile() {
     l=t stderr "Please see $2 for full details."
 }
 
-maybe_patch() {
```
jadeallenx, line 727: I think we should keep this function as a no-op because I feel certain we will run into a situation in the future where we need it due to some kind of shenanigans either at the Linux compiler level or upstream. (No shade intended for the OTP team, but mistakes get made and released sometimes.)

jadeallenx (changes requested):

kerl
paulo-ferraz-oliveira: Sure. I removed it since we could always recover it from source control, but I can put it back.

paulo-ferraz-oliveira: And is it just the function? No content? Or do we also restore the caller site?

jadeallenx: Yeah restore the call site too. I think passing in the params as-is is fine; it should be a no-op so won't hurt anything to supply them.

paulo-ferraz-oliveira: 8fbccc0

paulo-ferraz-oliveira: Lemme check where this is called, then...

paulo-ferraz-oliveira: 8fbccc0

jadeallenx (approved):

jadeallenx:
Looks good! Thanks for doing this!

paulo-ferraz-oliveira:
Squash-merging. I have a new one in the oven 👍

PR #469, opened 2023-10-02
Fix "`kerl deploy` not working"

### LINKED ISSUE #468: 'kerl deploy' not working
#### The bug

```
alex@hanzawa:~/src/kerl$ kerl deploy sarasti.local /home/alex/src/kerl/install/26
Couldn't ssh to sarasti.local
```

Using `strace` the problem is:
```
[pid 1186991] execve("/usr/bin/ssh", ["ssh", "", "sarasti.local", "true"], 0x5559e2a57be8 /* 43 vars */) = 0
```

Looking into the script the problem is that `$KERL_DEPLOY_SSH_OPTIONS` is double quoted which passes through.
```
if ! ssh "$KERL_DEPLOY_SSH_OPTIONS" "$host" true >/dev/null 2>&1; then
    l=e stderr "Couldn't ssh to $host"
    exit 1
fi  
```

Similarly for the `rsync` part:
```
if ! rsync -aqz -e "ssh $KERL_DEPLOY_SSH_OPTIONS" "$KERL_DEPLOY_RSYNC_OPTIONS" "$path/" "$host:$remotepath/"; then
    l=e stderr "Couldn't rsync Erlang/OTP $rel ($path) to $host ($remotepath)"
    exit 1
fi
```

Seeing this, we can work around this by using NOP values, the following works:
```
KERL_DEPLOY_SSH_OPTIONS='-6' KERL_DEPLOY_RSYNC_OPTIONS='-e ssh' kerl deploy sarasti.local ~/src/kerl/install/26
```

I would suggest dropping the double quotes as if the user provides multiple options it is going to break, but if they really are needed you probably can use instead:

```
${KERL_DEPLOY_SSH_OPTIONS:+"$KERL_DEPLOY_SSH_OPTIONS"}
```

#### Software versions

* Operating System: Debian 'bookworm' 12
* `kerl`: tip of `main`: `46ea061d4ce7db2ed1de3099ac0a36e31744fac4`

#### How to replicate

1. run `kerl deploy ...`

#### Expected behaviour

SSH/RSYNC over the OTP goodness.

#### Additional context

Any other context about the bug.

paulo-ferraz-oliveira:
# Description

When shellcheck'ing and shfmt'ing, [previously](https://github.com/kerl/kerl/pull/463), these ones got away.

Closes #468.

@jimdigriz, would you be able to test this and report back?

(you can see that there was an attempt to reverting previous chances by the fact that in one place the `disable` directive was kept when potentially unnecessary)

### DIFF
--- a/kerl
+++ b/kerl
@@ -1550,7 +1550,8 @@ do_deploy() {
         remotepath="$3"
     fi
 
-    if ! ssh "$KERL_DEPLOY_SSH_OPTIONS" "$host" true >/dev/null 2>&1; then
+    # shellcheck disable=SC2086  # Double quote to prevent globbing and word splitting
+    if ! ssh $KERL_DEPLOY_SSH_OPTIONS "$host" true >/dev/null 2>&1; then
         l=e stderr "Couldn't ssh to $host"
         exit 1
     fi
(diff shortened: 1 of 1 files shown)


paulo-ferraz-oliveira:
@jadeallenx, no need to tag or release, just for this, since this never got into a tag or release.

jadeallenx (approved):

jimdigriz:
@paulo-ferraz-oliveira Works For Me(tm)

paulo-ferraz-oliveira:
Thanks for catching this even before a new release 👍

PR #471, opened 2023-10-04
Try to make it easier to maintain for packages (and distros)

### LINKED ISSUE #431: Warn about non-existence of gcc-c++ (as an alternative to g++)
In some *nix systems, package `g++` (detected by `kerl` as a required dependency for functioning properly) might be available via `gcc-c++`. `kerl` should thus detect that either one or the other is present in the host system, since checking for `g++` alone, issue a warning and then work properly is misleading.

### LINKED ISSUE #439: Testing for package managers is getting to be a big mess
Proposed strategy: an associative array (i.e., a map in erlang terms, or a dict in python terms) of a command string with a label - execute the command and if it returns "true" then add the label to an output array. When we get the output array back we can see how many elements are in it and then call the appropriate testing functions (if any) for installed packages,

paulo-ferraz-oliveira:
# Description

The initial goal was to ease maintenance for "missing package" -related messages. Conceptually, this code is not very different from the previous one except that:

1. distros are dynamic (their ID is obtained from `os-releases` - with fallback) and then used to either "error" out or find the list of "probe-able" packages
2. the probes are, as before, unique for all packages, but since this is an evaluated command we can play with more complex probes
3. we now have distinct distros (this might actually, in a first release, reduce the number of targets we can use - since "package manager" is much more generic than "distro"), so we can also have e.g. multiple package managers per distro and allow for different package names (Fedora's `g++` is `gcc-c++` in RHEL - for the same package manager)
4. there's an environment variable to silence the output

## Further considerations

I had hacked the think locally to test with `darwin` and e.g. `rebar3` but reverted those changes before pushing.

~At the same time, ⚠️ I haven't tested in actual targets, but hope to do so with a few different Docker images.~

I'm testing now:

- [x] alpine
- [x] debian
- [x] fedora
- [x] linuxmint
- [ ] ~opensuse~
- [x] rhel
- [x] ubuntu

## Also take a look at

* [x] [Kerl depends on git and curl but it also depends on a bunch of other stuff...](https://github.com/kerl/kerl/pull/471#issuecomment-1748892642)
* [x] [If all logs add a prefix...](https://github.com/kerl/kerl/pull/471#issuecomment-1748967305)
* [x] ~[... the "package" checker output a command to install the missing packages...](https://github.com/kerl/kerl/pull/471#issuecomment-1752151295)~

Closes #439.
Closes #431.

### DIFF
--- a/kerl
+++ b/kerl
@@ -750,12 +750,12 @@ probe_pkgs() {
     os_release_id=$(get_id_from_os_release_files)
 
     if [ "${os_release_id}" = "1" ]; then
-        msg="[packages] Unable to determine Linux flavor (no release files); not checking build packages."
+        msg="[packages] Unable to determine Linux distro (no release files); not checking build packages."
         echo "${msg}" >>"$LOGFILE"
         l=w stderr "${msg}"
         return
     elif [ "${os_release_id}" = "2" ]; then
-        msg="[packages] Unable to determine Linux flavor (no ID); not checking build packages."
+        msg="[packages] Unable to determine Linux distro (no ID); not checking build packages."
         echo "${msg}" >>"$LOGFILE"
         l=w stderr "${msg}"
         return
@@ -778,7 +778,7 @@ probe_pkgs() {
             fi
         done
     else
-        msg="[packages] Unknown Linux flavor ${os_release_id}; not checking build packages."
+        msg="[packages] Unknown Linux distro ${os_release_id}; not checking build packages."
         echo "${msg}" >>"$LOGFILE"
         l=w stderr "${msg}"
     fi
--- a/kerl
+++ b/kerl
@@ -763,8 +763,7 @@ probe_pkgs() {
 
     kpp=$(eval echo \$_KPP_"${os_release_id}"_pkgs)
     if [ -n "${kpp}" ]; then
-        echo "[packages] Found package declarations for ${os_release_id}. Linux flavor is known." >>"$LOGFILE"
-        echo "[packages] Acting on known Linux distro ${os_release_id}" >>"$LOGFILE"
+        echo "[packages] Found package declarations for your Linux distro: ${os_release_id}" >>"$LOGFILE"
         for pkg in ${kpp}; do
             cmd=$(eval echo "\$_KPP_${os_release_id}_probe ${pkg}")
             probe=$(${cmd})
--- a/kerl
+++ b/kerl
@@ -135,7 +135,7 @@ KERL_GIT_DIR=${KERL_GIT_DIR:="${KERL_BASE_DIR:?}"/gits}
 KERL_GIT_BASE=https://raw.githubusercontent.com/kerl/kerl/master
 KERL_COLORIZE=${KERL_COLORIZE:=$KERL_COLOR_AVAILABLE}
 KERL_INCLUDE_RELEASE_CANDIDATES=${KERL_INCLUDE_RELEASE_CANDIDATES:=no}
-KERL_CHECK_BUILD_PACKAGES=${KERL_CHECK_BUILD_PACKAGES:="1"}
+KERL_CHECK_BUILD_PACKAGES=${KERL_CHECK_BUILD_PACKAGES:="yes"}
 
 if [ -n "$OTP_GITHUB_URL" ]; then
     _OGU="$OTP_GITHUB_URL"
@@ -818,7 +818,7 @@ _do_build() {
         # knows better than us and are going to go ahead and try to build
         # Erlang anyway. But at least there will be a clear warning to the
         # user if a build fails.
-        if [ "${KERL_CHECK_BUILD_PACKAGES}" = "1" ]; then
+        if [ "${KERL_CHECK_BUILD_PACKAGES}" = "yes" ]; then
             probe_pkgs
         fi
         ;;
--- a/kerl
+++ b/kerl
@@ -715,9 +715,9 @@ parse_os_release() {
     # returns:
     # - 2 if release ID is not found in os-release file
     os_release_id0=$(\grep "^ID=" "$1")
-    os_release_id=$(echo "${os_release_id0}" | \sed 's|.*=||')
+    os_release_id=$(echo "${os_release_id0}" | \sed 's|.*=||' | \sed 's|"||g')
     os_release_pretty_name0=$(\grep "^PRETTY_NAME=" "$1")
-    os_release_pretty_name=$(echo "${os_release_pretty_name0}" | \sed 's|.*=||')
+    os_release_pretty_name=$(echo "${os_release_pretty_name0}" | \sed 's|.*=||' | \sed 's|"||g')
     echo "[packages] Found ${os_release_pretty_name} with ID ${os_release_id}" >>"$LOGFILE"
     if [ -z "${os_release_id}" ]; then
         echo "[packages] Release ID not found in $1" >>"$LOGFILE"
--- a/kerl
+++ b/kerl
@@ -686,7 +686,7 @@ _apk() {
 # command, in variable _KPP_<distro>_probe to check from the package manager.
 # It should return 0 if package installation is Ok, non-0 otherwise.
 
-_KPP_alpine_pkgs="openssl-dev make autoconf ncurses-dev gcc g++"
+_KPP_alpine_pkgs="openssl-dev make automake autoconf ncurses-dev gcc g++"
 _KPP_alpine_probe="_apk"
 
 _KPP_debian_pkgs="libssl-dev make automake autoconf libncurses5-dev gcc g++"
--- a/kerl
+++ b/kerl
@@ -669,18 +669,15 @@ _flags() {
 }
 
 _dpkg() {
-    # gratefully stolen from
-    # https://superuser.com/questions/427318/test-if-a-package-is-installed-in-apt
-    # returns 0 (true) if found, 1 otherwise
-    echo "dpkg-query -Wf'\${db:Status-abbrev}' \"$1\" 2>/dev/null | \grep -q '^i'"
+    echo "dpkg-query -f='\${Status}' --show $1 | grep \"installed\""
 }
 
 _rpm() {
-    echo "rpm --quiet -q \"$1\" >/dev/null 2>&1"
+    echo "rpm -q \"$1\""
 }
 
 _apk() {
-    echo "apk -e info \"$1\" >/dev/null 2>&1"
+    echo "apk -e info \"$1\""
 }
 
 # To add package guessing magic for your Linux distro/package,
--- a/kerl
+++ b/kerl
@@ -135,6 +135,7 @@ KERL_GIT_DIR=${KERL_GIT_DIR:="${KERL_BASE_DIR:?}"/gits}
 KERL_GIT_BASE=https://raw.githubusercontent.com/kerl/kerl/master
 KERL_COLORIZE=${KERL_COLORIZE:=$KERL_COLOR_AVAILABLE}
 KERL_INCLUDE_RELEASE_CANDIDATES=${KERL_INCLUDE_RELEASE_CANDIDATES:=no}
+KERL_CHECK_BUILD_PACKAGES=${KERL_CHECK_BUILD_PACKAGES:="1"}
 
 if [ -n "$OTP_GITHUB_URL" ]; then
     _OGU="$OTP_GITHUB_URL"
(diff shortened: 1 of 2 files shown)


kerl
```diff
@@ -135,6 +135,7 @@ KERL_GIT_DIR=${KERL_GIT_DIR:="${KERL_BASE_DIR:?}"/gits}
 KERL_GIT_BASE=https://raw.githubusercontent.com/kerl/kerl/master
 KERL_COLORIZE=${KERL_COLORIZE:=$KERL_COLOR_AVAILABLE}
 KERL_INCLUDE_RELEASE_CANDIDATES=${KERL_INCLUDE_RELEASE_CANDIDATES:=no}
+KERL_CHECK_BUILD_PACKAGES=${KERL_CHECK_BUILD_PACKAGES:="1"}
```
paulo-ferraz-oliveira, line 138: Behaviour is kept, or do you prefer to change the default?

kerl
```diff
@@ -524,93 +525,6 @@ assert_build_name_unused() {
...
-    has_rpm=$(command -v rpm)
-    has_apk=$(command -v apk)
-    if [ -n "$has_dpkg" ] || [ -n "$has_rpm" ] || [ -n "$has_apk" ]; then
-        # found either dpkg, rpm or apk (or maybe even both!)
-        if [ -n "$has_dpkg" ] && [ -n "$has_rpm" ]; then
-            l=w stderr 'WARNING: You appear to have BOTH rpm and dpkg. This is very strange. No package checks done.'
-        elif [ -n "$has_dpkg" ] && [ -n "$has_apk" ]; then
-            l=w stderr 'WARNING: You appear to have BOTH apk and dpkg. This is very strange. No package checks done.'
-        elif [ -n "$has_rpm" ] && [ -n "$has_apk" ]; then
-            l=w stderr 'WARNING: You appear to have BOTH apk and rpm. This is very strange. No package checks done.'
```
paulo-ferraz-oliveira, line 538: We aren't doing this "multiple package managers" check any more.

kerl
```diff
@@ -754,7 +668,133 @@ _flags() {
...
+    dpkg-query -Wf'${db:Status-abbrev}' "$1" 2>/dev/null | \grep -q '^i'
+}
+
+_rpm() {
+    rpm --quiet -q "$1" >/dev/null 2>&1
+}
+
+_apk() {
+    apk -e info "$1" >/dev/null 2>&1
+}
```
paulo-ferraz-oliveira, line 684: This is basically the same as before.

kerl
```diff
@@ -754,7 +668,133 @@ _flags() {
...
+
+# To add package guessing magic for your Linux distro/package,
+# create a variable named _KPP_<distro>_pkgs where the content
+# is an array with packages, and then create a unique probe
+# command, in variable _KPP_<distro>_probe to check from the package manager.
+# It should return 0 if package installation is Ok, non-0 otherwise.
+# Special feature: the probe's \$1 is used for string substitution to eval it.
+
+_KPP_alpine="openssl-dev make autoconf ncurses-dev gcc g++"
+_KPP_alpine_probe="_apk \$1"
```
paulo-ferraz-oliveira, line 694: If we decide to go by package manager we have less entries, but basically go back to where we were, without the possibility to define different package names per different distro.

Also, I tried a bunch of stuff here, mostly associative arrays (which could give us granularity such as "a command per package", but ultimately this is a POSIX-sh, so not much we can do there, with arrays that is)...

kerl
```diff
@@ -754,7 +668,133 @@ _flags() {
...
+    else
+        echo "[packages] Release ID found in $1: ${os_release_id}" >>"$LOGFILE"
+        echo "${os_release_id}"
+    fi
+}
+
+get_id_from_os_release_files() {
+    # returns:
+    # 1 - if no release files exist
+    files="/etc/os-release /usr/lib/os-release"
```
paulo-ferraz-oliveira, line 735: Fallback is as per _standard_.

kerl
```diff
@@ -754,7 +668,133 @@ _flags() {
...
+    kpp=$(eval echo \$_KPP_"${os_release_id}"_pkgs)
+    if [ -n "${kpp}" ]; then
+        echo "[packages] Found package declarations for ${os_release_id}. Linux flavor is known." >>"$LOGFILE"
+        echo "[packages] Acting on known Linux distro ${os_release_id}" >>"$LOGFILE"
+        for pkg in ${kpp}; do
+            cmd=$(eval echo "\$_KPP_${os_release_id}_probe")
+            probe_without_1=$(echo "${cmd}" | sed "s|\$1||")
+            if [ "${probe_without_1}" != "${cmd}" ]; then
+                cmd=$(echo "${cmd}" | sed "s|\$1|${pkg}|")
+            fi
```
paulo-ferraz-oliveira, line 777: This is mentioned above, and gives us more flexibility. Since we're evaluating strings, I figured we need a way to call internal functions. This'll most likely be the regular use case.

kerl
```diff
@@ -754,7 +668,133 @@ _flags() {
...
+    else
+        msg="[packages] Unknown Linux flavor ${os_release_id}; not checking build packages."
+        echo "${msg}" >>"$LOGFILE"
+        l=w stderr "${msg}"
+    fi
+}
+
 _do_build() {
+    LOGFILE="$KERL_BUILD_DIR/$2/otp_build_$1.log"
+
```
paulo-ferraz-oliveira, line 797: Because `probe_pkgs` needs it 😄

paulo-ferraz-oliveira:
@jadeallenx, I'm starting tests, with Alpine as the first candidate. Out-of-the-box it doesn't have `curl` installed (at least not the Docker image I'm using). I know the README mentions Kerl depends on `git` and `curl` but it also depends on a bunch of other stuff, potentially existing in _all_ base distros, these days, like `grep`, `sed`, `cut`, `awk`, `find`, `cat`, `chmod`, `echo`, `printf`, `tail` (potentially more). I'm wondering if we should test for this list, or simply update the README to mention more than the 2 commands, there, or say we'll outright crash if a command isn't present. Thoughts?

paulo-ferraz-oliveira:
Might be worth mentioning all logs have a `[packages]` prefix.

---

## Idea 💡

If all logs add a prefix, too, we could even think of a way to filter out undesirable ones, e.g. 

```
ns="packages"
msg="Unable to determine"
log $ns $msg
```

and then inside log we could search for filters in a e.g. `KERL_LOG_FILTERS=` variable we'd expose.

(food for thought for another pull request 😄)

paulo-ferraz-oliveira:
Example logfile output

```
[packages] Found "Alpine Linux v3.18" with ID alpine
[packages] Release ID found in /etc/os-release: alpine
[packages] Found package declarations for alpine. Linux flavor is known.
[packages] Acting on known Linux distro alpine
[packages] Probe success for openssl-dev (distro: alpine)!
[packages] Probe success for make (distro: alpine)!
[packages] Probe success for autoconf (distro: alpine)!
[packages] Probe success for ncurses-dev (distro: alpine)!
[packages] Probe success for gcc (distro: alpine)!
[packages] Probe success for g++ (distro: alpine)!
```

kerl
paulo-ferraz-oliveira: @jadeallenx, I wonder if it'd make sense to reset to logfile to empty, at this point.

It was previously (and it still compatible with)

```
    LOGFILE="$KERL_BUILD_DIR/$2/otp_build_$1.log"
    # Add a build date and env in logfile
    echo "*** $(date) - kerl build $1 ***" >>"$LOGFILE"
```

but this means we have several build logs in this file, forever growing. Maybe it's intended (?)

kerl
```diff
@@ -754,7 +668,135 @@ _flags() {
...
+# To add package guessing magic for your Linux distro/package,
+# create a variable named _KPP_<distro>_pkgs where the content
+# is an array with packages, and then create a unique probe
+# command, in variable _KPP_<distro>_probe to check from the package manager.
+# It should return 0 if package installation is Ok, non-0 otherwise.
+# Special feature: the probe's \$1 is used for string substitution to eval it.
+
+_KPP_alpine_pkgs="openssl-dev make autoconf ncurses-dev gcc g++"
+_KPP_alpine_probe="_apk \$1"
```
paulo-ferraz-oliveira, line 694: If we decide to go by package manager we have less entries, but basically go back to where we were, without the possibility to define different package names per different distro.

Also, I tried a bunch of stuff here, mostly associative arrays (which could give us granularity such as "a command per package", but ultimately this is a POSIX-sh, so not much we can do there, with arrays that is)...

kerl
```diff
@@ -754,7 +668,135 @@ _flags() {
...
+        msg="[packages] Unknown Linux flavor ${os_release_id}; not checking build packages."
+        echo "${msg}" >>"$LOGFILE"
+        l=w stderr "${msg}"
+    fi
+}
+
 _do_build() {
+    LOGFILE="$KERL_BUILD_DIR/$2/otp_build_$1.log"
+    # Add a build date and env in logfile
+    echo "*** $(date) - kerl build $1 ***" >>"$LOGFILE"
```
paulo-ferraz-oliveira, line 798: Because `probe_pkgs` needs it 😄

paulo-ferraz-oliveira: @jadeallenx, I wonder if it'd make sense to reset to logfile to empty, at this point.

It was previously (and it still compatible with)

```
    LOGFILE="$KERL_BUILD_DIR/$2/otp_build_$1.log"
    # Add a build date and env in logfile
    echo "*** $(date) - kerl build $1 ***" >>"$LOGFILE"
```

but this means we have several build logs in this file, forever growing. Maybe it's intended (?)

paulo-ferraz-oliveira:
Can't get `alpine` to build properly. Keep getting 

```
make[2]: *** [aarch64-unknown-linux-musl/Makefile:704: aarch64-unknown-linux-musl/erl_db_insert_list.ycf.h] Killed
```

I looked at https://github.com/bitwalker/alpine-erlang/blob/master/Dockerfile, for example, and don't think I'm doing anything different from what they're doing. I installed all the packages that image install, and am passing `--build` as a configuration option.

kerl
```diff
@@ -754,7 +668,131 @@ _flags() {
...
+    echo "dpkg-query -Wf'${db:Status-abbrev}' \"$1\" 2>/dev/null | \grep -q '^i'"
+}
+
+_rpm() {
+    echo "rpm --quiet -q \"$1\" >/dev/null 2>&1"
+}
+
+_apk() {
+    echo "apk -e info \"$1\" >/dev/null 2>&1"
+}
```
paulo-ferraz-oliveira, line 684: This is basically the same as before. But we prepare the command, via an `echo` we further `eval`. Because when the probe fails I wanna show how we're probing...

kerl
```diff
@@ -754,7 +668,131 @@ _flags() {
...
+
+# To add package guessing magic for your Linux distro/package,
+# create a variable named _KPP_<distro>_pkgs where the content
+# is an array with packages, and then create a unique probe
+# command, in variable _KPP_<distro>_probe to check from the package manager.
+# It should return 0 if package installation is Ok, non-0 otherwise.
+
+_KPP_alpine_pkgs="openssl-dev make autoconf ncurses-dev gcc g++"
+_KPP_alpine_probe="_apk"
```
paulo-ferraz-oliveira, line 693: If we decide to go by package manager we have less entries, but basically go back to where we were, without the possibility to define different package names per different distro.

Also, I tried a bunch of stuff here, mostly associative arrays (which could give us granularity such as "a command per package", but ultimately this is a POSIX-sh, so not much we can do there, with arrays that is)...

kerl
```diff
@@ -754,7 +668,131 @@ _flags() {
...
+    echo "dpkg-query -Wf'\${db:Status-abbrev}' \"$1\" 2>/dev/null | \grep -q '^i'"
+}
+
+_rpm() {
+    echo "rpm --quiet -q \"$1\" >/dev/null 2>&1"
+}
+
+_apk() {
+    echo "apk -e info \"$1\" >/dev/null 2>&1"
+}
```
paulo-ferraz-oliveira, line 684: This is basically the same as before. But we prepare the command, via an `echo` we further `eval`. Because when the probe fails I wanna show how we're probing...

kerl
paulo-ferraz-oliveira: Ran into this already, while testing `rhel`. Package for `g++` is `gcc-c++`. Don't know if it's the same for Fedora, for example...

paulo-ferraz-oliveira:
## Idea 💡

This is maybe outside the scope of `kerl` but how would you feel having the "package" checker output a command to install the missing packages? (this would, of course, be a different pr, to avoid "idea" creep).

paulo-ferraz-oliveira:
As a bonus, and to check if it'd be easy to extend to e.g. Arch Linux, using `pacman` I'm also testing a local version for that. If I got it right, I only need a probe function and some package names...

jadeallenx:
> ## Idea 💡
> This is maybe outside the scope of `kerl` but how would you feel having the "package" checker output a command to install the missing packages? (this would, of course, be a different pr, to avoid "idea" creep).

Hm. This sounds like a lot of maintenance work. To be honest, I don't really want kerl to do _any_ package detection but when builds break, it is typically because build prerequisites are not available and then people open tickets that "kerl" is broken somehow. So all of this effort is to help people figure out that it's not kerl's fault lol - not sure we need to help people that much?

What benefit do you see if this were an option?

kerl
jadeallenx: I'd prefer if the default were "yes" or "true" rather than 1

kerl
```diff
@@ -754,7 +668,128 @@ _flags() {
...
+        msg="[packages] Unable to determine Linux flavor (no ID); not checking build packages."
+        echo "${msg}" >>"$LOGFILE"
+        l=w stderr "${msg}"
+        return
+    fi
+
+    kpp=$(eval echo \$_KPP_"${os_release_id}"_pkgs)
+    if [ -n "${kpp}" ]; then
+        echo "[packages] Found package declarations for ${os_release_id}. Linux flavor is known." >>"$LOGFILE"
+        echo "[packages] Acting on known Linux distro ${os_release_id}" >>"$LOGFILE"
```
jadeallenx, line 767: Seems like you could combine these messages.

`[packages] Your linux seems to be ${os_release_id}.` ???

kerl
jadeallenx: Hm. Feel like the entire process should be in the same log file... maybe we should define LOGFILE as "otp_build_$1.log" earlier

jadeallenx (changes requested):

kerl
paulo-ferraz-oliveira: I'll change it. 👍 I did think about it, at a point in time, but while looking at the other variables, found that `KERL_COLORIZE`, and `KERL_AUTOCLEAN` both mention `0`/`1`, and not `no`/`yes`.

paulo-ferraz-oliveira: Granted `KERL_INCLUDE_RELEASE_CANDIDATES` has `no` as a default... 😄

paulo-ferraz-oliveira: Granted `KERL_INCLUDE_RELEASE_CANDIDATES` has `no` as a default... 😄

paulo-ferraz-oliveira: 6505b69

kerl
paulo-ferraz-oliveira: Yeah, I guess so. A previous implementation had them separated into different code paths, but one I joined it into this single `if` didn't take care of it. 👍

paulo-ferraz-oliveira: 7fa76c3, then f6deb2b to not have it called "flavor" sometimes and "distro" some other times 😄

kerl
paulo-ferraz-oliveira: What do you mean "earlier"? Earlier than `_do_build()`? I can change that in the scope of this pull request, but it feel like scope creep 😄

paulo-ferraz-oliveira: Earlier than this, in either `do_normal_build` or `do_git_build` there's only validations, which I'm not sure warrant a log entry, but lemme know...

paulo-ferraz-oliveira:
> What benefit do you see if this were an option?

Apart from copy-paste ease, none, so I agree we shouldn't do it.

paulo-ferraz-oliveira:
I did the request changes. Based on our prior discussions for this stuff, do you think the end result is close to what you expected? Or do you think it'll be a lot of maintenance work to keep adding distros/packages? (I do understand that `kerl` could do without all this, but it's also - as you marginally state - part of its value as a tool)

jadeallenx:
> should test for this list, or simply update the README to mention more than the 2 commands, there, or say we'll outright crash if a command isn't present. Thoughts?

We've survived this long noting only git and curl, I think we can rely on those other utilities being part of the base install tbqh

kerl
jadeallenx: Yeah we should drop them or stick to one consistent word :)

kerl
jadeallenx: OK I guess I'm confused then - I think we _should_ log the package probing output along with all of the build time spam from compilers and what not. I don't think we need to log validations.

jadeallenx (approved):
lgtm

jadeallenx:
OK don't kill me but seeing all of the

```bash
echo "${msg}" >> $LOGFILE
```

makes me think we should make a function called log() which takes a string and prepends the date and writes it to a filehandle

jadeallenx:
Eh. On second thought maybe we should open a new issue/PR for log emission cleanup/generalization - because stderr() is a specialized log emitter than writes to stderr and not a disk file.

jadeallenx:
> Might be worth mentioning all logs have a `[packages]` prefix.
> 
> ## Idea 💡
> If all logs add a prefix, too, we could even think of a way to filter out undesirable ones, e.g.
> 
> ```
> ns="packages"
> msg="Unable to determine"
> log $ns $msg
> ```
> 
> and then inside log we could search for filters in a e.g. `KERL_LOG_FILTERS=` variable we'd expose.
> 
> (food for thought for another pull request 😄)

lol great minds - yeah we should definitely open an issue to track this - don't think this work needs to delay a release tho?

paulo-ferraz-oliveira:
There we go: https://github.com/kerl/kerl/issues/475

paulo-ferraz-oliveira:
I'm squash-merging this in preparation for the release. I think we can hold off on it for a few more days (some consumers use the main branch) to see if this surfaces any "issues", but we don't have to if you don't wanna.

jadeallenx:
Yeah let's plan for a release on Thursday

PR #480, opened 2023-10-19
Simplify code maintenance

### LINKED ISSUE #194: Replace env var crunching with neat xargs trick

paulo-ferraz-oliveira:
# Description

Eases maintenance by auto-assigning variables declared as strings in `_KERL_CFGS`.

Closes #194.

### DIFF
--- a/kerl
+++ b/kerl
@@ -225,7 +225,7 @@ KERL_INCLUDE_RELEASE_CANDIDATES=${KERL_INCLUDE_RELEASE_CANDIDATES:=no}
 KERL_CHECK_BUILD_PACKAGES=${KERL_CHECK_BUILD_PACKAGES:="yes"}
 
 act_on_kerl_cfgs() {
-    # $1: kerl config. flags
+    # $1: kerl config. flags (we're only interested in the keys)
     # $2: action - cache | restore (default)
 
     for _KERL_CFG in $1; do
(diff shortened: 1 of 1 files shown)


kerl
```diff
@@ -224,115 +224,66 @@ KERL_COLORIZE=${KERL_COLORIZE:=$KERL_COLOR_AVAILABLE}
...
+KERL_INSTALL_MANPAGES= \
+KERL_INSTALL_HTMLDOCS= \
+KERL_BUILD_PLT= \
+KERL_BUILD_DOCS= \
+KERL_DOC_TARGETS=chunks \
+KERL_BUILD_BACKEND= \
+KERL_RELEASE_TARGET= \
+"
+
+save_default_kerl_cfgs() {
```
paulo-ferraz-oliveira, line 248: I'm looking at this function now, and thinking I could get away with a simple `eval $_KERL_CFGS`.

paulo-ferraz-oliveira:
If you write

```bash
            echo "$_KERL_TARGET_VAR=\"$_KERL_CFG_VAR\""
```

just before the last `eval` on `act_on_kerl_cfgs` you can see it in action 😄

kerl
```diff
@@ -281,6 +261,9 @@ KERL_BUILD_DOCS=
 KERL_DOC_TARGETS=chunks
 KERL_BUILD_BACKEND=
 KERL_RELEASE_TARGET=
+"
+act_on_kerl_cfgs "$_KERL_CFGS" "cache"
```
paulo-ferraz-oliveira, line 265: Called it here to have the `_KERL_CFGS` variable defined, but it does (similar to) what's above, from lines 227-268.

PR #481, opened 2023-10-21
Fix `is_older_than_x_days`: use `find` instead of `stat`

paulo-ferraz-oliveira:
# Description

I'm moving to `find`, which seems to be more closely related in FreeBSD, Ubuntu and macOS, for example. (`stat` was also failing in Ubuntu, as it was only tested on macOS).

### DIFF
--- a/kerl
+++ b/kerl
@@ -616,9 +616,8 @@ is_older_than_x_days() {
     # $1: file to check
     # $2: age in days
 
-    eval "$(\stat -s "$1")"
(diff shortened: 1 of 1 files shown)


paulo-ferraz-oliveira:
I had previously tested this only on macOS, but today I launched a Ubuntu container to start on a new pr and found this issue.

jadeallenx (approved):

PR #485, opened 2023-10-26
Warn on stale build due to kernel changed

### LINKED ISSUE #207: Kerl fails to install previous builds if kernel upgraded
This is due to the `otp_release` script looking in the wrong directory for a makefile.  The error message kerl give you is terrible - it says "failed to install" but there's no reason for the failure.

So we should try to make this failure more clear to the end user and one way of doing that is to check to see if the kernel has been upgraded versus when the build was done.  We could put the `uname -r` output into our build metadata file and check that for example.

paulo-ferraz-oliveira:
# Description

Closes #207.

### DIFF
--- a/kerl
+++ b/kerl
@@ -1075,6 +1075,7 @@ _do_build() {
     TMPOPT="${TMP_DIR}/kerloptions.$$"
     {
         echo "This is kerl's control file for build configuration."
+        echo "Please don't edit it manually!"
         echo "CFLAGS: ${CFLAGS:-}"
         echo "KERL_CONFIGURE_OPTIONS: $KERL_CONFIGURE_OPTIONS"
         echo "$(uname_r_label): $(uname_r)"
--- a/kerl
+++ b/kerl
@@ -990,6 +990,14 @@ fail_do_build() {
     fi
 }
 
+uname_r_label() {
+    echo "uname -r"
+}
+
+uname_r() {
+    eval "$(uname_r_label)"
+}
+
 _do_build() {
     init_build_logfile "$1" "$2"
     log_build_entry "*** $(date) - kerl build $1 ***"
@@ -1069,7 +1077,7 @@ _do_build() {
         echo "This is kerl's control file for build configuration."
         echo "CFLAGS: ${CFLAGS:-}"
         echo "KERL_CONFIGURE_OPTIONS: $KERL_CONFIGURE_OPTIONS"
-        echo "uname -r: $(uname -r)"
+        echo "$(uname_r_label): $(uname_r)"
     } >>"$TMPOPT"
     SUM=$($MD5SUM "$TMPOPT" | cut -d ' ' -f "$MD5SUM_FIELD")
     # Check for a .kerl_config.md5 file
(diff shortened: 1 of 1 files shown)


kerl
```diff
@@ -1078,8 +1087,8 @@ _do_build() {
         fi
     else
         # no file exists, so write one
-        mv "$TMPOPT" .kerl_config
-        echo "$SUM" >.kerl_config.md5
+        mv "$TMPOPT" "$KERL_CONFIG_STORAGE_FILENAME"
+        echo "$SUM" >"$KERL_CONFIG_STORAGE_FILENAME".md5
```
paulo-ferraz-oliveira, line 1091: Variable already existed. I just took advantage of that, for consistency.

kerl
```diff
@@ -1618,6 +1627,16 @@ do_install() {
     ERL_TOP="$KERL_BUILD_DIR/$build_name/otp_src_$rel"
     cd "$ERL_TOP" || exit_install
     init_install_logfile "$rel" "$build_name"
+
+    prev_build_kernel_release=$(grep <"$ERL_TOP"/"$KERL_CONFIG_STORAGE_FILENAME" -o '^uname -r: \(.*\)$' | sed -n 's|^uname -r: \(.*\)$|\1|p')
```
paulo-ferraz-oliveira, line 1631: This is now "far" from the place where we put the result of `uname -r` in the file; should we change it to a function, or add a comment where the source is created, for example?

jadeallenx: Probably should be a function

paulo-ferraz-oliveira: Perfect. I'll change that.

paulo-ferraz-oliveira: 6372d76: now `uname -r` (the function) and "uname -r" (the label) are tightly bound to the code that uses them (for creation and validation)

kerl
```diff
@@ -985,6 +990,14 @@ fail_do_build() {
     fi
 }
 
+uname_r_label() {
+    echo "uname -r"
```
jadeallenx, line 994: Are we sure this is the right command on all of our platforms?? probably is - but for md5 and sed we had to detect platform before we used a command like this...

kerl
```diff
@@ -1060,8 +1073,12 @@ _do_build() {
 
     # Check to see if configuration options need to be stored or have changed
     TMPOPT="${TMP_DIR}/kerloptions.$$"
-    echo "${CFLAGS:-}" >"$TMPOPT"
-    echo "$KERL_CONFIGURE_OPTIONS" >>"$TMPOPT"
+    {
+        echo "This is kerl's control file for build configuration."
```
jadeallenx, line 1077: I would add something like "please don't edit it manually"

kerl
```diff
@@ -1618,6 +1635,16 @@ do_install() {
...
     init_install_logfile "$rel" "$build_name"
+
+    prev_build_kernel_release=$(grep <"$ERL_TOP"/"$KERL_CONFIG_STORAGE_FILENAME" -o "^$(uname_r_label): \(.*\)\$" | sed -n "s|^$(uname_r_label): \(.*\)\$|\1|p")
+    if [ "$(uname_r)" != "$prev_build_kernel_release" ]; then
+        warn "this Erlang/OTP build appears to be stale. It was created with kernel release"
+        warn "         '$prev_build_kernel_release' while currently your system's kernel release is" "noprefix"
+        warn "         '$(uname_r)'." "noprefix"
+        warn "         You should consider removing the build with 'kerl delete build ...' and" "noprefix"
+        warn "         re-installing it with 'kerl build-install ...'" "noprefix"
+    fi
```
jadeallenx, line 1646: This is great

kerl
paulo-ferraz-oliveira: I can check further. At the same time, I'm not sure:

1. we're always using `\` (sometimes we are) where commands are executed
  1.1. I can later pull request for this, maybe (?)
2. we're using the `sed` options consistently (don't seem to be)
  2.1. I can later pull request for this, maybe (?)

paulo-ferraz-oliveira: `uname` should be POSIX, and `-r` is defined as "Write the current release level of the operating system implementation."

paulo-ferraz-oliveira: From what I could gather, `uname -r` is valid for Darwin, OpenBSD, FreeBSD, and a host of Linux systems.

jadeallenx: > 1. we're always using `\` (sometimes we are) where commands are executed

I don't remember the exact commit where it was introduced (think it was some kind of shellcheck clean up???) but it is supposed to make it so the command is executed in its own environment without environment variables or something?

>    1.1. I can later pull request for this, maybe (?)

Yeah maybe? low priority tho

jadeallenx (approved):

kerl
paulo-ferraz-oliveira: > I don't remember the exact commit where it was introduced

ShellCheck does not complain about this sort of stuff, AFAIK. If I remember correctly is was an actual bug somebody wanted to fix...

> Yeah maybe? low priority tho

I have no real priorities (I'm kinda going, with `kerl`, with oldest issues first - if that's any type of priority order), but these should mostly be mechanical changes 😄

paulo-ferraz-oliveira:
I'm squash+merging... Thanks for the review.

kerl
paulo-ferraz-oliveira: Ref: https://github.com/kerl/kerl/pull/363

PR #491, opened 2023-11-09
Be more specific in the packages we suggest for installation

### LINKED ISSUE #249: automake and autoconf only required when git is the build backend
They are spurious warning messages otherwise.

paulo-ferraz-oliveira:
# Description

Only suggest `automake` and `autoconf` when `KERL_BUILD_BACKEND` is `git`.

Closes #249.

## Further considerations

This is at the cost of a bit more maintenance, though most likely the software lists aren't bound to change much any time soon.

### DIFF
--- a/kerl
+++ b/kerl
@@ -873,46 +873,32 @@ _apk() {
     echo "apk -e info \"$1\""
 }
 
+common_ALL_pkgs="gcc make"
+common_ALL_BUILD_BACKEND_git_pkgs="automake autoconf"
+common_debian_pkgs="${common_ALL_pkgs} libssl-dev libncurses5-dev g++"
+
 # To add package guessing magic for your Linux distro/package,
 # create a variable named _KPP_<distro>_pkgs where the content
 # is an array with packages, and then create a unique probe
 # command, in variable _KPP_<distro>_probe to check from the package manager.
 # It should return 0 if package installation is Ok, non-0 otherwise.
 
-_KPP_alpine_pkgs="openssl-dev make ncurses-dev gcc g++"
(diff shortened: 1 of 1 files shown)


kerl
```diff
@@ -879,22 +879,40 @@ _apk() {
...
-_KPP_rhel_pkgs="openssl-devel make automake autoconf ncurses-devel gcc gcc-c++"
+_KPP_rhel_pkgs="openssl-devel make ncurses-devel gcc gcc-c++"
+if [ "$KERL_BUILD_BACKEND" = 'git' ]; then
+    _KPP_rhel_pkgs="${_KPP_rhel_pkgs} automake autoconf"
+fi
 _KPP_rhel_probe="_rpm"
 
-_KPP_ubuntu_pkgs="libssl-dev make automake autoconf libncurses5-dev gcc g++"
+_KPP_ubuntu_pkgs="libssl-dev make libncurses5-dev gcc g++"
+if [ "$KERL_BUILD_BACKEND" = 'git' ]; then
```
jadeallenx, line 913: seems like this could be a common check for all linux flavors?

paulo-ferraz-oliveira: Yeah, I thought about it. We could do the same exercise for other packages, but then I thought "We decided to keep them separate before, so...". I don't mind separating them, but, in that case, I'll try to aggregate as much of the common stuff as possible.

jadeallenx: i think it makes sense to collapse common things into a common code path

paulo-ferraz-oliveira: 0bc39e5

jadeallenx (approved):

PR #508, opened 2024-02-18
fix: `activate` breaks `man` for non-OTP man pages

### LINKED ISSUE #498: `activate` breaks `man` for non-OTP man pages
#### The bug

After `activate`ing an OTP installation, `man` not longer finds manpages.

The problem is that `activate` (for all shell) sets the `MANPATH` environment variable. A non-existing `MANPATH` means `use the system default`. Setting that variable overrides the `man` build-in defaults, disabling them.
The `activate` scripts should check if `MANPATH` is set, and if not, include the output of `manpath -g` in the newly set variable.
 
#### Software versions

* Ubuntu 23.10
* Erlang/OTP: 26.2.1
* `kerl`: 4.0

#### How to replicate

1. check that man works, e.g. `man sh`
2. activate an OTP install
3. `. /opt/otp-26.1/activate`
4. try the man page from step 1 again, will result in `No manual entry for sh`

#### Expected behaviour

man works after activating an installation.

zardoz03:
# Description
man by default from what i can tell checks /etc/manpath.config, which then breaks normal usage of man if MANPATH is set without this

fixes #498 

  might equally be `manpath -g`, however `manpath`, in my case, preserves user paths such as `$HOME/.local/share/man` 

### DIFF
--- a/kerl
+++ b/kerl
@@ -1314,7 +1314,7 @@ if [ -n "\${MANPATH+x}" ]; then
     if [ -n "\$MANPATH" ]; then
         MANPATH="\${_KERL_MANPATH_REMOVABLE}:\$MANPATH"
     else
-        MANPATH="\${_KERL_MANPATH_REMOVABLE}"
+        MANPATH="\${_KERL_MANPATH_REMOVABLE}:\$(manpath)"
     fi
 else
     MANPATH="\${_KERL_MANPATH_REMOVABLE}"
--- a/kerl
+++ b/kerl
@@ -1456,6 +1456,7 @@ set -x _KERL_MANPATH_REMOVABLE "$directory/lib/erlang/man" "$directory/man"
 set -p _KERL_CLEANUP "set -e _KERL_MANPATH_REMOVABLE;"
 if set -q MANPATH
     if test -n "\$MANPATH"
+        set -xp MANPATH "\$(manpath)"
         set -xp MANPATH \$_KERL_MANPATH_REMOVABLE
     else
         set -x MANPATH \$_KERL_MANPATH_REMOVABLE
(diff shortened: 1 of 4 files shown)


paulo-ferraz-oliveira:
👋 thanks for the pull request.

Regarding "[?] I have performed a self-review of my changes", I'm not sure the ? is because you don't understand the choice item, but I'll explain:
- after you pushed your changes, you made sure (in GitHub, since that's what we'll use to review initially) you pushed exactly what you want

paulo-ferraz-oliveira:
Tests failing...

paulo-ferraz-oliveira:
Are these changes consistent in the fish and C shell activation scripts?

paulo-ferraz-oliveira:
`manpath -g` does not seem to work in macOS, so `manpath` alone seems like an Ok solution.

zardoz03:
i think its because the CI checks the manpath after activation, and now there is the docker state included...

paulo-ferraz-oliveira:
> also sorry i neglected the csh/fish activation

No prob. We'd eventually either do it, or, at the very least, identify it as an issue 😄, but since you updated it, it's most welcome.

As per CI you might need to run `shellcheck tests/activate_test.sh` or maybe update the file as per the CI results and re-push.

paulo-ferraz-oliveira:
Would it make this all simpler if we:

1. fetch MANPATH with `manpath`
2. update a variable (maybe we don't need to check if it's set or not)
3. do cleanup upon `deactivate`

?

I'd need to test this better, maybe I'm missing a detail, but the current implementation seems much more complex than required 😕 @jadeallenx, thoughts?

zardoz03:
idk how i'd alter the tests, but the new failures seem to be that the expected_env is presumed to be the same on different calls/users, so it'd have to use something like `grep $EXPECTED` over the actual MANPATH, rather than literal equality

paulo-ferraz-oliveira:
> idk how i'd alter the tests, but the new failures seem to be that the expected_env is presumed to be the same on different calls/users, so it'd have to use something like `grep $EXPECTED` over the actual MANPATH, rather than literal equality

You could potentially also validate that it's, as expected, the concatenation you build.

paulo-ferraz-oliveira:
> idk how i'd alter the tests

You mean you don't know where to start, or what to do?

You can look at the `tests` folder for examples, or run your own, by writing a Bash script (`exit` with `1` on error, otherwise `0`).

paulo-ferraz-oliveira:
Closing as stale (also, all CI is failing). Feel free to re-open and apply the required changes.

PR #511, opened 2024-02-25
Extend `KERL_AUTOCLEAN` to Git-based builds

paulo-ferraz-oliveira:
# Description

`KERL_AUTOCLEAN` was being taken into account only for "normal" builds. We extend that behaviour to Git based builds.

At the same time we fix an issue were the logfile was said to be kept, but was actually removed upon failure.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,10 @@
 ---
 name: CI
-'on': [push, pull_request]
+'on':
+  - push
+concurrency:
+  group: ${{github.workflow}}-${{github.ref}}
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
paulo-ferraz-oliveira, line 1: These changes are not strictly required in this pull request's scope, but they do affect it.

.github/workflows/lint.yml
paulo-ferraz-oliveira, line 1: These changes are not strictly required in this pull request's scope, but they do affect it.

jadeallenx (approved):

PR #507, opened 2024-02-18
Prepare for OTP 27

paulo-ferraz-oliveira:
# Description

This can/should be accepted/merged (only) when 27 is moved from release candidate status.

⚠️ ~Opening as draft to signal that.~
⚠️ ~On second thought, there's an actual "issue" that might need solving, which is to use `ex_doc` to generate local documentation from 27 on.~ (this has been addressed, next to OTP, by @garazdawi, as mentioned in the linked issue)

**Note**: I'll try to remember to leave this here up until 27 is released, and then I'll release. If anything important comes up or is merged, I'll rebase.

## Reference

* https://erlangforums.com/t/erlang-otp-27-0-rc1-released/3300

- [x] I have performed a self-review of my changes
- [x] I have read and understood the [contributing guidelines](/kerl/kerl/blob/main/CONTRIBUTING.md)

PR #516, opened 2024-03-12
Have CI run on pull request, workflow dispatch and push to main

paulo-ferraz-oliveira:
# Description

A small change (as I'd done in #507) to allow CI to run under proper conditions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,7 +1,13 @@
 ---
 name: CI
 'on':
-  - push
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - "*"
+  workflow_dispatch: {}
 concurrency:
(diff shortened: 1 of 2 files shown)


paulo-ferraz-oliveira:
@jadeallenx, I'm adding you as a reviewer, for FYI, but if this goes through I'm merging (directly) to the main branch, because:

- a similar change had been approved for #507, already
- we're otherwise preventing CI from running in pull requests, which is not desired

paulo-ferraz-oliveira:
This is failing CI on macOS. I haven't dug deep into that, but it seems related to Python, on which `fop` depends. As it (`fop`) will not be a necessity in the future (given the doc -related changes in 27), and since it was there only to prevent a warning, not a real issue, I'm removing it.

jadeallenx (approved):

PR #523, opened 2024-05-06
feat: standardize on libncurses-dev for dpkg package probing

### LINKED ISSUE #522: Over-optimized for happy path in "Probe failed for libncurses5-dev" warning.

I think I overoptimized PR #521 for #518, the real intent of the warning ends up looking messy for users:

image

I'd like to give the fix another go with a more sophisticated approach that tests for either package, and then warns if none are found by favoring `libncurses-dev`.

My proposal is to include both libraries individually in `common_debian_pkgs` and then extend the probe loop by special-casing the `*ncurses*` package tests, only warn if none are found, and ensure this works for all os releases, not just deb-based.

I've tried to use expansion to test for either package, which works in a bash/zsh shell:

image

But not in sh:

image

#### Additional context

This will then encourage users to favor the "new" package, but not complain if the system only has "5" installed.

03juan:
# Description

Improve the error message introduced by relaxing the pkg probe with wildcards.

Closes #522 

## Improvement

```sh
$ ./kerl build 26.2.5
Extracting source code for normal build...
Building (normal) Erlang/OTP 26.2.5 (26.2.5); please wait...
Initializing (build) log file at /home/juan/.kerl/builds/26.2.5/otp_build_26.2.5.log.
WARNING: [packages] Probe failed for libncurses-dev (distro: ubuntu): probe "dpkg-query -Wf'${db:Status-abbrev}' "libncurses-dev" 2>/dev/null | \grep -q '^i'" returned 1
^C

$ dpkg-query -Wf'${db:Status-abbrev}' libncurses-dev libncurses5-dev 
dpkg-query: no packages found matching libncurses-dev
dpkg-query: no packages found matching libncurses5-dev

$ head -13 /home/juan/.kerl/builds/26.2.5/otp_build_26.2.5.log
*** Mon May  6 11:47:48 SAST 2024 - kerl build 26.2.5 ***
Build options:
* KERL_BUILD_DOCS=yes
[packages] Found Ubuntu 24.04 LTS with ID ubuntu
[packages] Release ID found in /etc/os-release: ubuntu
[packages] Found package declarations for your Linux distro: ubuntu
[packages] Probe success for automake (distro: ubuntu)!
[packages] Probe success for autoconf (distro: ubuntu)!
[packages] Probe success for gcc (distro: ubuntu)!
[packages] Probe success for make (distro: ubuntu)!
[packages] Probe success for libssl-dev (distro: ubuntu)!
[packages] Probe success for g++ (distro: ubuntu)!
[packages] Probe failed for libncurses-dev (distro: ubuntu): probe "dpkg-query -Wf'${db:Status-abbrev}' "libncurses-dev" 2>/dev/null | \grep -q '^i'" returned 1
```

### DIFF
--- a/kerl
+++ b/kerl
@@ -882,7 +882,7 @@ _apk() {
 
 common_ALL_pkgs="gcc make"
 common_ALL_BUILD_BACKEND_git_pkgs="automake autoconf"
-common_debian_pkgs="${common_ALL_pkgs} libssl-dev libncurses[^w]*dev g++"
+common_debian_pkgs="${common_ALL_pkgs} libssl-dev libncurses-dev g++"
 
 # To add package guessing magic for your Linux distro/package,
 # create a variable named _KPP_<distro>_pkgs where the content


kerl
```diff
@@ -971,20 +971,40 @@ probe_pkgs() {
...
             probe_res=$?
-            if [ "${probe_res}" != 0 ]; then
-                msg="[packages] Probe failed for ${pkg} (distro: ${os_release_id}): probe \"${probe}\" returned ${probe_res}"
-                log_build_entry "${msg}"
-                warn "${msg}"
-            else
-                log_build_entry "[packages] Probe success for ${pkg} (distro: ${os_release_id})!"
-            fi
+            case "${pkg}" in
+            *ncurses*)
```
03juan, line 984: This ensures `ncurses-dev` apk and `ncurses-devel` rpm are also considered so we also skip the test at L1004 for those os's.

paulo-ferraz-oliveira:
@03juan, thanks for the pr.

I'm not a big fan of the approach, though. That code will have to be maintained, and it's recently been "improved" to make probing simpler. Making it less so reverts us back to where we were.

Can we just go back to a specific (more recent) version name and apply the strategy that was in place? If you're using an older package/OS you get a warning, but that's about it, the process doesn't stop nor are you forced to take the warning into account.

Thanks.

03juan:
Completely agree

paulo-ferraz-oliveira:
I think you may need to rebase this on the main branch.

03juan:
My apologies. Rebased, squashed and force-pushed.

03juan:
Will you take care of the changelog entry?

paulo-ferraz-oliveira:
Yeah, no worries for the CHANGELOG. Thanks for the pr.

PR #527, opened 2024-06-03
Allow specifying Git clone depth on `kerl build git ...`

### LINKED ISSUE #447: support clone with depth 1
most of the time, when we build from a repo that we don't need to clone all history of this repo, this repo is temporary, hence we could clone it only with depth 1 to speed up 
bildo
with depth 1, the size can reduce to half

paulo-ferraz-oliveira:
# Description

... via `KERL_GIT_CLONE_DEPTH`, as documented in the README

We want splitting to occur, thus the ShellCheck disable directive

Closes #447.

- [ ] I have performed a self-review of my changes
- [ ] I have read and understood the [contributing guidelines](/kerl/kerl/blob/main/CONTRIBUTING.md)

### DIFF
--- a/README.md
+++ b/README.md
@@ -489,6 +489,13 @@ Available variables:
 
 The format of the prompt section to add.
 
+#### `KERL_GIT_CLONE_DEPTH`
+
+Default: `not set`
+Value passed to `git clone --depth=` when using `kerl build git ...`.
+
+**Note**: this doesn't always result in smaller clones.
+
 ### Installation configuration
 
 #### `KERL_DEFAULT_INSTALL_DIR`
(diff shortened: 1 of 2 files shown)


paulo-ferraz-oliveira:
@jadeallenx, I didn't add tests as this seems fairly innocuous (and if the current tests pass, it should indicate the current behaviour is kept - at least nothing's breaking with the new option), but lemme know if you'd prefer some.

jadeallenx (approved):

jadeallenx:
Looks good, thanks!

paulo-ferraz-oliveira:

PR #528, opened 2024-06-12
On macOS, use OpenSSL 1.1 if Erlang/OTP is pre-25.1, and OpenSSL 3.0 otherwise

### LINKED ISSUE #472: OpenSSL proposed as 1.1, on macOS with Homebrew installed

Not a "problem" but something to think about...

At the moment, if the consumer doesn't specify OpenSSL (e.g. `--with-ssl`), we try to guess it using Homebrew (macOS builds)...

```sh
        if ! echo "$KERL_CONFIGURE_OPTIONS" | \grep 'with-ssl' >/dev/null 2>&1; then
            whichbrew=$(command -v brew)
            if [ -n "$whichbrew" ] && [ -x "$whichbrew" ]; then
                brew_prefix=$(brew --prefix openssl@1.1)
                if [ -n "$brew_prefix" ] && [ -d "$brew_prefix" ]; then
                    KERL_CONFIGURE_OPTIONS="$KERL_CONFIGURE_OPTIONS "--with-ssl=$brew_prefix
                fi
            fi
        fi
```

We seem to have version 1.1 hardcoded, but Erlang/OTP is OpenSSL 3.0 -compatible since at least 25.1...

Adding info. to the README regarding what versions of OpenSSL Erlang/OTP supports (maybe there's a list out there somewhere, that we could use).

paulo-ferraz-oliveira:
# Description

Since Erlang/OTP 25.1, OpenSSL 3.0 is supported. We can leverage that for the default `build` behaviour, even if you can override it via a config. option.

Closes #472.

There's already a bit, in CI, that tests SSL, via `erl -s crypto -s init stop`.
At the same time, CI's not running for pre 25.1 (**edit**: since 25.0 is inside our chosen 25-27 range, I'm specifically adding 25.0 to CI - even if we decide to remove it before merge), so we can just "trust" that part of the code.

## Further considerations

Here's CI showing it:

* choosing OpenSSL 1.1 for Erlang/OTP 25.0: https://github.com/kerl/kerl/actions/runs/9490113882/job/26152873359?pr=528#step:6:1439
* choosing OpenSSL 3.0 for Erlang/OTP 25.3: https://github.com/kerl/kerl/actions/runs/9490113882/job/26152873809?pr=528#step:6:1485

The main difference between the current implementation and the previous one is that we're "forcing" 3.0 for 25.1+, which would before be forced as 1.1; this means that users that don't have 3.0 installed via brew will get a warning about it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,7 +30,7 @@ jobs:
             /bin/bash -c "$(curl -fsSL \
               https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
             HOMEBREW_NO_INSTALLED_DEPENDENTS_CHECK=yes \
-              brew install automake wxwidgets fish shellcheck shfmt coreutils
+              brew install automake wxwidgets fish shellcheck shfmt coreutils openssl@3.0
           else
             sudo sed -i 's/azure\.//' /etc/apt/sources.list # Reduces chance of time-outs
             sudo apt-get update -y
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,7 +21,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        otp_vsn: [25, 26, 27, master]
+        otp_vsn: ['25.0', '25', '26', '27', 'master']
         os: [ubuntu-22.04, macos-13]
     steps:
       - name: Update env.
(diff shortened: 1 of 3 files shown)


jadeallenx (approved):
This is great, thanks!

paulo-ferraz-oliveira:
Thank you. Squashing and merging.

paulo-ferraz-oliveira:

PR #532, opened 2024-06-25
Fix warning

t-hg:
# Description

Fixes the following warning:

```shell
WARNING: [packages] Probe failed for g++ (distro: fedora): probe "rpm -q "g++"" returned 1
```

As it can be seen "g++" is only an alias for "gcc-c++":

```shell
sudo dnf install g++
Last metadata expiration check: 3:38:57 ago on Tue 25 Jun 2024 11:09:27 CEST.
Package gcc-c++-14.1.1-6.fc40.x86_64 is already installed.
Dependencies resolved.
Nothing to do.
Complete!
```

- [x] I have performed a self-review of my changes
- [ ] I have read and understood the [contributing guidelines](/kerl/kerl/blob/main/CONTRIBUTING.md)
   **No**, because link is broken

PR #534, opened 2024-06-25
Have shared code reusable (on the package warning list)

paulo-ferraz-oliveira:
# Description

After I merged #532, I got to thinking that we'd done what I did now, before, but for Debian. So I repeat it for "a little less maintenance".


jadeallenx (approved):

paulo-ferraz-oliveira:

PR #535, opened 2024-06-26
Fix kerl path in activate script generation

ruilapa:
# Description

OTP activate script can point to a non existent kerl.

Example scenarios:
- OTP built on a different path/system
- brew kerl path changed

Closes #&lt;issue&gt;.

- [X] I have performed a self-review of my changes
- [X] I have read and understood the [contributing guidelines](/kerl/kerl/blob/master/CONTRIBUTING.md)

### DIFF
--- a/kerl
+++ b/kerl
@@ -1334,7 +1334,7 @@ if type kerl_deactivate 2>/dev/null | \grep -qw function; then
     kerl_deactivate
 fi
 
-_KERL=$(command -v kerl)
+_KERL=\$(command -v kerl)
 if [ -n "\$_KERL" ]; then
     _KERL_VERSION=\$(\$_KERL version)
 fi
@@ -1596,7 +1596,7 @@ if ( \$status == 0 ) then
     kerl_deactivate
 endif
 
-set _KERL = $(command -v kerl)
+set _KERL = \$(command -v kerl)
 
 if ( "\$_KERL" != "" ) then
   set _KERL_VERSION = \`\$_KERL version\`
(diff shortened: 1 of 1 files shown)


paulo-ferraz-oliveira:
Look, it's @ruilapa (@jadeallenx, we used to work together before 😄)

Lemme look at this.

paulo-ferraz-oliveira (approved):
Change is reasonable. If this passes CI (which tests the activation scripts) it'll get merged.

paulo-ferraz-oliveira:
My Bash fails me, it seems. You can emit the activation scripts with [`emit-activate`](https://github.com/kerl/kerl?tab=readme-ov-file#emit-activate) and then check/lint the result to make sure what you fixed is actually what you need fixed (remember this is not the activation script itself; this is running in the context of writing another script from a script).

paulo-ferraz-oliveira (changes requested):
CI's failing...

paulo-ferraz-oliveira:
Actually, what is possible you might also do is emit the activation script and replace with the previous one. Also, to be complete, the same change should be done to all activation scripts: `bash`, `csh`, and `fish`.

paulo-ferraz-oliveira:
Re-running CI... there's lint issues: https://github.com/kerl/kerl/actions/runs/9700785319/job/26779695039?pr=535

paulo-ferraz-oliveira:
This breaks CI and no update was proposed. Closing... feel free to re-open if you have an alternative. Thanks.

PR #537, opened 2024-06-26
Force `CI` env. variable even if not set by GitHub Actions

paulo-ferraz-oliveira:
# Description

It's a fairly common practice (that the CI environment sets this variable), and we're taking advantage of it in `kerl`'s code, but it seems GitHub Actions might not always honor this "tradition", so we force it. I'm doing this to help in debugging https://github.com/kerl/kerl/pull/535, because I see CI, fail but no message ~, when we should at least see "Colorization disabled as 'tput' (via 'ncurses') seems to be unavailable." issued from `kerl` itself~ (read the code wrong, it's a `-z` there).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,6 +14,7 @@ concurrency:
 env:
   ERLC_USE_SERVER: true
   KERL_DEBUG: 'yes'
+  CI: true
 jobs:
   ci:
     name: CI OTP ${{matrix.otp_vsn}}, on ${{matrix.os}}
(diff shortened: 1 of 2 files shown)


paulo-ferraz-oliveira:
Let's wait on this one, I'm not convinced I'm reading the error wrong. We're doing `echo "$match"`, line 2539, which means we get a path, but then the `activate` script fails, which is why we have `exit code 1`.

paulo-ferraz-oliveira:
Nah, `CI` is always set to `true`, as per https://docs.github.com/en/actions/learn-github-actions/variables. And now we have a reference. For the issue in the other pull request I'm gonna try to understand it using a local Ubuntu container.

PR #539, opened 2024-07-03
fix: kerl upgrade always upgrading

### LINKED ISSUE #538: kerl upgrade always upgrading
#### The bug

`kerl upgrade` is always triggering an  upgrade, regardless of the current version.

#### Software versions

* Operating System: macOS 13.5
* `kerl`: 4.1.1

#### How to replicate

1. run `kerl upgrade`
2. run `kerl upgrade` again

#### Expected behaviour

At least the second step shouldn't trigger an upgrade.

#### Additional context

This is happening because the local version isn't properly placed. This was introduced in [8c3c7a31490678b88810d75368c7ad7782c604f](https://github.com/kerl/kerl/commit/f8c3c7a31490678b88810d75368c7ad7782c604f) when the output  of `kerl version` (and the other commands) was being sent to stderr, making the assignment `version=$(kerl version)` to be empty.

In https://github.com/Alberdi/kerl/commit/a3a1ef671188db155aa6b523e093d81f508ec1c2 I've tried a solution that seems to work, and I could create a PR or iterate over it (with my limited shell skills). I'm not sure if changing the `tip` call to an `echo` is the proper solution, but that would also help with external integrations that might want to parse the kerl version number.

Alberdi:
# Description

I've noticed that calling `kerl upgrade` was always triggering an upgrade because the local version wasn't being properly set. This was introduced in f8c3c7a31490678b88810d75368c7ad7782c604f when the output of `kerl version` (and the other commands) was being sent to stderr, making the assignment `version=$(kerl version)` to be empty.

I'm not sure if changing the `tip` call to an `echo` is the proper solution, but this would also help with external integrations that might also want to parse the kerl version number, and that's what the included test is testing.

Closes #538.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -132,6 +132,8 @@ jobs:
         run: tests/activate_test.fish $_KERL_REL $_KERL_VSN $(./kerl path install_"$_KERL_VSN")
       - name: Test activate/cleanup (csh)
         run: tests/activate_test.csh $_KERL_REL $_KERL_VSN $(./kerl path install_"$_KERL_VSN")
+      - name: Test version parse (sh/bash)
+        run: tests/version_parse_test.sh
       - name: Delete installation
         run: ./kerl delete installation $(./kerl path install_"$_KERL_VSN")
       - name: Delete build
--- a/kerl
+++ b/kerl
@@ -2330,7 +2330,7 @@ version)
         usage_exit "version"
     fi
 
-    tip "$KERL_VERSION"
+    echo "$KERL_VERSION"
     ;;
 build)
     if [ "$2" = 'git' ]; then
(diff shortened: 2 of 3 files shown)


paulo-ferraz-oliveira:
Tests are failing. It's possible you either have to:

1. change the path to `kerl`, or
2. activate `kerl`, or
3. rely on another step's output to call `kerl` from the correct place

Alberdi:
> Tests are failing. It's possible you either have to:
> 
> 1. change the path to `kerl`, or
> 2. activate `kerl`, or
> 3. rely on another step's output to call `kerl` from the correct place

I see, sorry about that. I'll review this and also the listing step.

paulo-ferraz-oliveira:
No worries. That's what CI's there for 😄 The change is sane, though, and after CI's ✅ I'll test it locally and then eventually squash+merge+release.

paulo-ferraz-oliveira:

paulo-ferraz-oliveira:
Thanks for this, @Alberdi.

PR #541, opened 2024-07-04
Fix `kerl upgrade` example on README.md

Alberdi:
# Description

The text didn't match the expected (and current) script behaviour, because it was saying that the old version was still available.

Closes #&lt;issue&gt;.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,10 +60,10 @@ Run:
 
 ```console
 $ kerl upgrade
-Local kerl found (/usr/local/bin/kerl) at version 3.0.0.
-Remote kerl found at version 4.2.0.
-Versions are different. Upgrading to 4.2.0...
-kerl 4.2.0 is now available at /usr/local/bin/kerl.
+Local kerl found (/usr/local/bin/kerl) at version 2.6.0.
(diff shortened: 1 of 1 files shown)


paulo-ferraz-oliveira:
👋 thanks for the pull request, but I don't think we need to update it. It's documentation, as an example (as you state); we just put that there kinda for the format.

Alberdi:
I just wanted to point out that the format wasn't right. In case it was not clear with all the numbers changing, before the PR the text was:
```
Local kerl found (/usr/local/bin/kerl) at version $OLD_VERSION.
Remote kerl found at version $NEW_VERSION.
Versions are different. Upgrading to $NEW_VERSION...
kerl $OLD_VERSION is now available at /usr/local/bin/kerl.
```

So I've changed that last line in the example to reference `$NEW_VERSION` instead, as the `kerl` script properly does.

Even then, it's a very minor thing.

paulo-ferraz-oliveira:
Oh, I see.

paulo-ferraz-oliveira:
Perfect. Thanks for this.

PR #552, opened 2024-12-30
on OpenBSD we need gtar instead of tar

DavidAlphaFox:
# Description
On OpenBSD, the `tar` command is different with GNU/Linux's `tar`. So there is a command under `/usr/local/bin` named `gtar`。The `gtar` command can work with well with the given parameters in the `kerl`.
So I add a  `case` statement to test the `KERL_SYSTEM` variable to decide which 'tar' program should be used. 

Closes #&lt;issue&gt;.

### DIFF
--- a/kerl
+++ b/kerl
@@ -468,8 +468,14 @@ _curl() {
 
 unpack() {
     # $1: <file> to extract to current directory
-
-    tar -x --no-same-owner --no-same-permissions -zf "$1"
+    case "$KERL_SYSTEM" in
+	OpenBSD)
+    	gtar -x --no-same-owner --no-same-permissions -zf "$1"
+	;;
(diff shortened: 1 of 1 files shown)


paulo-ferraz-oliveira:
👋 are these commands to not be compatible? If it's `gtar`'s goal to be `tar`-compatible, for example, you could further reduce your code by just having the executable be the variable element; otherwise I think it's Ok as-is.

jadeallenx:
Quibbles:

1. Don't the other BSDs need gtar too?
2. I'd prefer all of the operating system configuration happen in the block [here](https://github.com/kerl/kerl/blob/master/kerl#L307-L318) with all of the other differences between GNU style systems and BSDs

PR #566, opened 2026-02-28
Fix MANPATH assignment in kerl script

### LINKED ISSUE #498: `activate` breaks `man` for non-OTP man pages

schnef:
If `MANPATH` is empty, man will look in the standard location `/usr/shar/man/...` In the script `activate`, if `MANPATH` is empty, `MANPATH` is set to `_KERL_MANPATH_REMOVABLE` which sets `MANPATH` to non-empty and that's why `man` wont find the man pages anymore. By adding a single `:` to the end of the `MANPATH` when it is empty will add an empty path and fix the problem.

# Description

Added empty paths to `MANPATH` if `MANPATH` is empty to start with and is replaced by `_KERL_MANPATH_REMOVABLE`

Closes #498.

### DIFF
--- a/kerl
+++ b/kerl
@@ -1404,10 +1404,10 @@ if [ -n "\${MANPATH+x}" ]; then
     if [ -n "\$MANPATH" ]; then
         MANPATH="\${_KERL_MANPATH_REMOVABLE}:\$MANPATH"
     else
-        MANPATH="\${_KERL_MANPATH_REMOVABLE}"
+        MANPATH="\${_KERL_MANPATH_REMOVABLE}:"
     fi
 else
-    MANPATH="\${_KERL_MANPATH_REMOVABLE}"
+    MANPATH="\${_KERL_MANPATH_REMOVABLE}:"
     add_cleanup "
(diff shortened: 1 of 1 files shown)


paulo-ferraz-oliveira:
Thanks. When CI's Ok, I'll merge and release.

